PR #8, opened 2021-12-01
Changes requested by DWA Group:

krishmaj:
- the component testing history entries now show the test times as well as dates
- added the ability to query a single Wire Tension value on the Wire Tension test results page
- improved the layout of the Wire Tension test results page and plots within, and fixed a bug where the wire numbers were being displayed on half-integer values
- fixed a bug with incorrect links to components from within the component and test type forms
- fixed a small code formatting inconsistency in the edit_component.pug file
- removed the explicit list of bugs from the homepage, and changed to now direct the user to the DB's Github issues pages

### DIFF
--- a/pug/home.pug
+++ b/pug/home.pug
@@ -57,10 +57,13 @@ block content
   pre #{''}
   
   div.border-danger.border.rounded.p-2
-    h5 Known Bugs:
+    h5 Bugs and Other Issues
     ul
-      li All pages - the page titles are not showing up in browser tabs
-      li All component pages - the component history is unavailable (null date error)
+      li Bugs and other errors with existing functionality should be reported at our 
+         |
+         a(href = "https://github.com/DUNE/dunedb/issues") Github Issues page
+         |  (a Github account is required).  Please give as much information as you can - an example page or link where the bug occurs, how to reproduce the error (if appropriate), etc.
+      li Requests for new features and functionality can also be made on that page, but please first make sure that the addition cannot instead be made by users through the database interface.
       
   pre #{''}
   pre #{''}
--- a/pug/home.pug
+++ b/pug/home.pug
@@ -61,7 +61,7 @@ block content
     ul
       li Bugs and other errors with existing functionality should be reported at our 
          |
-         a(href = "https://github.com/DUNE/dunedb/issues") Github Issues page
+         a(href = "https://github.com/DUNE/dunedb/issues", target = "_blank") Github Issues page
          |  (a Github account is required).  Please give as much information as you can - an example page or link where the bug occurs, how to reproduce the error (if appropriate), etc.
       li Requests for new features and functionality can also be made on that page, but please first make sure that the addition cannot instead be made by users through the database interface.
       
--- a/pug/test_viewResult.pug
+++ b/pug/test_viewResult.pug
@@ -46,7 +46,7 @@ block extrascripts
       var side = document.getElementById("apaSide").value;
       var number = document.getElementById("wireNumber").value;
       
-      var keyName = "wires." + layer + "." + side + "[" + number + "]";
+      var keyName = `wires.${layer}.${side}[${number}]`;
       document.getElementById("tensionKey").value = keyName;
       
       var tension = testdata.data.wires[layer][side][number];
--- a/pug/test_viewResult.pug
+++ b/pug/test_viewResult.pug
@@ -37,6 +37,21 @@ block extrascripts
       form.nosubmit = true;
       form.submission = testdata;
     }
+  
+  script(type = 'text/javascript').
+    
+    function GetWireTension()
+    {
+      var layer = document.getElementById("wireLayer").value;
+      var side = document.getElementById("apaSide").value;
+      var number = document.getElementById("wireNumber").value;
+      
+      var keyName = "wires." + layer + "." + side + "[" + number + "]";
+      document.getElementById("tensionKey").value = keyName;
+      
+      var tension = testdata.data.wires[layer][side][number];
+      document.getElementById("wireTension").value = tension;
+    };
 
 block content
   .container-fluid
@@ -71,7 +86,39 @@ block content
       else
         dt span.alert.alert-danger Component information is not available!
     
-    pre #{''}
+      pre #{''}
+        
+      if(formrec.formName == "Wire Tensions")
+        .row
+          .col-9
+            div.border-warning.border.rounded.p-2
+              .row
+                .col <strong>Query a Wire Tension: </strong>
+                .col
+                  select.form-control(id = "wireLayer", type = "text")
+                    option(value = '' ) Wire Layer
+                    option(value = 'X') X
+                    option(value = 'U') U
+                    option(value = 'V') V
+                    option(value = 'G') G
+                .col
+                  select.form-control(id = "apaSide", type = "text")
+                    option(value = '' ) APA Side
+                    option(value = 'A') A
+                    option(value = 'B') B
+                .col
+                  input.form-control(id = "wireNumber", type = "text", placeholder = "Wire Number")
+                .col
+                  input.btn.btn-info#submitPromotion(type = "submit", onclick = "GetWireTension()")
+                .col-2
+                  input.form-control(id = "tensionKey", type = "text", placeholder = "Tension Key", readonly)
+                .col-2
+                  input.form-control(id = "wireTension", type = "text", placeholder = "Wire Tension (N)", readonly)
+          
+          .col-3
+      
+        pre #{''}
+        pre #{''}
     
     div.border-success.border.rounded.p-2
       #builtform  
@@ -111,7 +158,6 @@ block content
           =process.state
           |  - #{process.created.length} objects created
 
-
     p
       if(permissions.userHas(user,'testdatas:process'))
         each process, processId in (formrec.processes || {})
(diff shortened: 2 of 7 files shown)


pug/home.pug
```diff
@@ -57,10 +57,13 @@ block content
...
   
   div.border-danger.border.rounded.p-2
-    h5 Known Bugs:
+    h5 Bugs and Other Issues
     ul
-      li All pages - the page titles are not showing up in browser tabs
-      li All component pages - the component history is unavailable (null date error)
+      li Bugs and other errors with existing functionality should be reported at our 
+         |
+         a(href = "https://github.com/DUNE/dunedb/issues") Github Issues page
```
micchickenburger: Consider whether adding `target = "_blank"` would be better to avoid navigating away from the page if the user needs to report a bug.

pug/test_viewResult.pug
```diff
@@ -37,6 +37,21 @@ block extrascripts
...
+  
+  script(type = 'text/javascript').
+    
+    function GetWireTension()
+    {
+      var layer = document.getElementById("wireLayer").value;
+      var side = document.getElementById("apaSide").value;
+      var number = document.getElementById("wireNumber").value;
+      
+      var keyName = "wires." + layer + "." + side + "[" + number + "]";
```
micchickenburger: Just as an FYI, you can use string interpolation instead of concatenation:
```
const keyName = `wires.${layer}.${side}[${number}]`;
```

pug/test_viewResult.pug
```diff
@@ -71,7 +86,39 @@ block content
...
+                  input.btn.btn-info#submitPromotion(type = "submit", onclick = "GetWireTension()")
+                .col-2
+                  input.form-control(id = "tensionKey", type = "text", placeholder = "Tension Key", readonly)
+                .col-2
+                  input.form-control(id = "wireTension", type = "text", placeholder = "Wire Tension (N)", readonly)
+          
+          .col-3
+      
+        pre #{''}
+        pre #{''}
```
micchickenburger, line 121: Why are preformatted elements used here?

pug/test_viewResult.pug
```diff
@@ -132,3 +178,6 @@ block content
     div
       a.btn.btn-info(href = '/test/' + testdata._id + '/copyToDraft') Copy Test to New Draft
   
+    pre #{''}
+    pre #{''}
```
micchickenburger, line 182: Why are preformatted elements used here?

static/components/ComponentArray.js
```diff
@@ -58,25 +58,21 @@ class ArrayComponent extends TextFieldComponent{
...
+          tpl += `<div>Max. Tension (N): <span class='arrayComponentMax'></span></div>`;
+          tpl += `<div>Mean Tension (N): <span class='arrayComponentMean'></span></div>`;
+          tpl += `<div>RMS Tensions (N): <span class='arrayComponentRMS'></span></div>`;
+          var bounds = this.getBounds();
+          if(!isNaN(bounds.hi)) tpl += `<div>Out-of-spec high: <span class='arrayComponentOobHi'>n/a</span></div>`;
+          if(!isNaN(bounds.lo)) tpl += `<div>Out-of-spec low: <span class='arrayComponentOobLo'>n/a</span></div>`;
+        tpl += `</div>`
+        tpl += `<div class='flex-grow-1 p-4 arrayComponentGraph' style='height: 200px; width: 240px;'></div>`;
+        tpl += `<div class='flex-grow-1 p-4 arrayComponentHistogram' style='height: 200px; width: 240px;'></div>`;
+      tpl += `</div>`
```
micchickenburger: Consider using multiline string interpolation instead of concatenation for better readability:
```
tpl = `
  <div>Max: ... </div>
  <div>Mean: ... </div>
  ...
  <div>
    <span>Number of Wires...</span>
  </div>
`;
```

static/components/ComponentArray.js
```diff
@@ -237,11 +232,6 @@ class ArrayComponent extends TextFieldComponent{
     console.log("this.value",this.value);
     // debugger;
```
micchickenburger, line 215: I know this isn't your code but it would be a good idea to remove all commented code and console statements from the code.  If we need to reference deleted code, we can look at the file history in git.  It's just a code hygiene thing.  One big question everyone reading the code will have is "why is this code commented out?"

static/components/ComponentUUID_Component.js
```diff
@@ -126,7 +126,7 @@ class ComponentUUID_Component extends QR_Component{
     // }
     if(this.refs.compUuidInfo && value && value.length==36) {
```
micchickenburger, line 127: I don't know if you know the answer to this, but `value.length==36` begs a comment for explanation.

micchickenburger (dismissed):

pug/test_viewResult.pug
krishmaj: I actually tried this exactly as you've written, and it didn't work for some reason.

pug/test_viewResult.pug
krishmaj: I've found that it's the best way to just get some spacing between elements and at the bottom of pages ... most of the other methods that I've tried don't consistently work for whatever reason.

static/components/ComponentArray.js
krishmaj: Is on my todo list actually - exactly as you've said, to remove all unused code and logger/console lines.  Wanted to get more of the DB checked over first before doing that though.

static/components/ComponentUUID_Component.js
krishmaj: No idea to be honest.  Can investigate if you want.

pug/home.pug
krishmaj: Will do.

pug/test_viewResult.pug
krishmaj: As above.

pug/test_viewResult.pug
micchickenburger: [Broken Windows Theory](https://en.wikipedia.org/wiki/Broken_windows_theory), though appropriately criticized in the psychology and criminology fields where it originated, definitely applies to software engineering.  Hacking things to work achieves the desired outcome but perpetuates more hacks in the future as contributors care less and less about code quality.  And of course code quality is directly tied to maintainability (poor code quality creates technical debt, which manifests as increased cost in time and resources for writing bug-free code because existing code is unclear, undocumented, or buggy.)

So in this case `pre` elements are only intended to be used to render preformatted text.  To use them for any other purpose creates technical debt.  Sometimes there are no obvious solutions for a problem, in which case a hack may be necessary.  But if that's the case the hack must be clearly documented:
* This is the problem we face
* This is the solution we chose and why
* These are the expected inputs
* This is the expected output

If the problem is adding spacing, the solution is to use CSS.  CSS is the solution to every style-related issue.  HTML should *only* markup data (e.g. this is a table of data, this is a header, this is a paragraph, etc.)  Colors, borders, outlines, padding, margins, backgrounds, animations, shapes, etc--these are all within the realm of CSS.

pug/test_viewResult.pug
micchickenburger: If the code gets transpiled, the PUG transpiler might not support ESNext, in which case we might just need to update it.  Natively, however, all modern browsers support this feature.

krishmaj: To be fair, I don't actually know why it wasn't working ... I'm aware that string interpolation should work with my browser, so I knew it wasn't that which was the issue.  I'll give it another go ... might have been something else on the page that was breaking things (I was playing with other things on it at the same time).

pug/test_viewResult.pug
krishmaj: Fair enough.  Now, baring in mind that I've done the exact same thing in every other pug file that I've looked over in the past 2 months, can we leave this here for this commit and I will put in a separate commit where I scrub all of the pre statements in favour of the correct code?  Would like to have some time to try out what works and what doesn't, rather than rushing a change for this commit.

static/components/ComponentArray.js
krishmaj: Good point - was mainly just trying to keep as close to the original code as I could, but why keep it difficult.

pug/test_viewResult.pug
micchickenburger: Sure, I think that's fine.  Creating an issue for this would be a good idea too so others would know there is intention to remove them.  :)

static/components/ComponentArray.js
krishmaj: I'll make an issue for this one as well, again so that people know it's coming.

static/components/ComponentUUID_Component.js
krishmaj: The 36 refers to the total length (number of characters) in the UUID pattern: 8 + 4 + 4 + 4+ 12, plus 4 dashes.

static/components/ComponentArray.js
krishmaj: I've changed as much as I could without breaking the formatting ... but I'm sure it can be improved further (I'm not yet so good with combining string interpolation and HTML tags), so I'm happy to make further changes.

micchickenburger: I cleaned up that file a bit more, stylistically.  See how the string interpolation can be accomplished with variables?  In the case of conditionals, ternary expressions can be used.
image

Issue #10, opened 2021-12-01
Remove commented out code and console/logger statements


krishmaj:
I know this isn't your code but it would be a good idea to remove all commented code and console statements from the code.  If we need to reference deleted code, we can look at the file history in git.  It's just a code hygiene thing.  One big question everyone reading the code will have is "why is this code commented out?"

_Originally posted by @micchickenburger in https://github.com/DUNE/dunedb/pull/8#discussion_r760462611_

krishmaj:
Various commits over the past few months have slowly removed unneeded and unused code, comments and console / logger statements.  PR #91 made the final cleanup, so this issue is now resolved.

PR #11, opened 2021-12-26
Overhaul of the Job entity code:

krishmaj:
- made the naming consistent ... workflows were always the same thing as jobs, and now all under the latter
- cleaned up the backend library and routes code for jobs (partially addressing issue #10) - removed unused functions and code, console and logger statements, and added better comments
- cleaned up and improved layouts on frontend pug pages, and renamed files for consistency
- moved Job History into the main Job Information viewing page, since there is generally so little in the former that its not worth having a separate page for it
- removed the ability to make draft jobs, in favour of allowing jobs to be edited (like components) ... for right now, this seems more appropriate, but can re-address in the future if needed
- renamed the Test Information viewing pug file to be consistent with the equivalent Component and Job ones


krishmaj:
@micchickenburger Verification is still not working FYI - as before, I wasn't asked for any passphrase on my end while creating the commit.

krishmaj:
Forgot to mention in my commit message - there are 2 functions in the `Jobs.js` file: `search` and `save` which I have not cleaned up.  These are used exclusively by the API side, which I have absolutely no idea about yet, so I don't want to make changes without a way to test things out afterwards.

micchickenburger:
> @micchickenburger Verification is still not working FYI - as before, I wasn't asked for any passphrase on my end while creating the commit.

I would go over the steps again in my blog post:  https://www.micah.soy/posts/setting-up-git-identities/.  In particular, the global config setting `user.useConfigOnly` should be set to `true`.  This will stop any commits without a configured identity.  Also make sure any global identity is unset per step one.

pug/edit_jobTypeForm.pug
```diff
@@ -0,0 +1,198 @@
...
+
+
+block metaschema
+  
+  // Set up the layout of the various entry boxes, using the 'edit_form.pug' template as a initial starting point
+  script(type = 'text/javascript').
+    
+    var builder_wizard = false;
+
+    var metaschema = {"components": [{"label": "Columns",
```
micchickenburger: This indentation feels inconsistent and hard to read.  Typical formatting in Javascript is like so:
```javascript
let someArray = [{
  key: [{
    foo: 'bar',
    baz: 'foo',
  }, {
    foo: 'bar',
    baz: 'foo',
  }],
}];
```

pug/job.pug
```diff
@@ -0,0 +1,98 @@
...
+    
+    // Get information about the job and its corresponding type form
+    var job = !{JSON.stringify(job || {}, null, 4)};
+    var form = !{JSON.stringify(form || {}, null, 4)};
+  
+  script(type = 'text/javascript').
+    
+    // Get the job's type form, and populate it with the job's actual submission
+    // Then disable the submission button (since the form is only to be displayed, not used)
+    window.onload = async function()
```
micchickenburger: This will break any other window onload function that might exist.  Use `window.addEventListener('load', function);` instead.

pug/job.pug
```diff
@@ -0,0 +1,98 @@
...
+  - var page_title = "View Submitted Job";
+
+
+block extrascripts
+  
+  script(type = 'text/javascript').
+    
+    // Get information about the job and its corresponding type form
+    var job = !{JSON.stringify(job || {}, null, 4)};
+    var form = !{JSON.stringify(form || {}, null, 4)};
```
micchickenburger, line 16: I don’t recognize this syntax… it looks like casting from string to boolean.  Is this a pug interpolation feature of some kind?  Maybe a comment here would be useful to describe what’s happening.

pug/job_submit.pug
```diff
@@ -0,0 +1,138 @@
...
+  - var page_title = "Submit Job";
+
+
+block extrascripts
+  
+  script(type = 'text/javascript').
+    
+    // Get information about the job and corresponding type form
+    var job = !{JSON.stringify(job || {}, null, 4)}; 
+    var form = !{JSON.stringify(form || {}, null, 4)};
```
micchickenburger, line 15: Same as above…  What is `!{}` supposed to do?

pug/job_submit.pug
```diff
@@ -0,0 +1,138 @@
...
+            
+          // Go to the page for viewing information about the submitted job, and do not leave the submission page in the browser history
+          window.location.replace('/job/' + retval);
+        }
+        
+        // If the submission fails ...
+        function postFail(res, statusCode, statusMsg)
+        {
+          // Set up and relay error messages
+          if(res.responseText && res.responseText.length > 0)
```
micchickenburger: These conditions are redundant.  Empty string is falsey in JavaScript.

pug/job_submit.pug
```diff
@@ -0,0 +1,138 @@
...
+          {
+            jobForm.setAlert("danger", res.responseText);
+          }
+          else
+          {
```
micchickenburger, line 56: String interpolation could be used here to improve readability.

pug/list_jobTypes.pug
```diff
@@ -0,0 +1,123 @@
...
+      
+      script.
+        $('button.NewJobTypeButton').click(function()
+        {
+          window.location.href = '/jobs/' + $('button.NewJobTypeButton').parents('.modal-content').find('input').val() + '/new';
+        })
+        
+        $('input.NewJobTypeName').on("keyup",function(e)
+        {
+          if(e.key == "Enter")
```
micchickenburger: Should always use strict compare `===` to avoid automatic type casting

pug/job_submit.pug
```diff
@@ -0,0 +1,138 @@
...
+                method: "post",
+                url: '/json/job',
+                data: JSON.stringify(submission),
+                dataType: "json",
+                success: postSuccess})
+         .fail(postFail);
+      }
+      catch(e)
+      {
+        // TOFIX: not sure what to put here (used to be console and debugger, but we're no longer using them)
```
micchickenburger, line 74: If we don’t have a logging solution of any kind maybe we should remove the try…catch statement altogether.  Removing the statement will allow errors to throw, which will be visible in the console but may also break something in the UI.  Leaving the catch in hides the errors.

routes/draftRoutes.js
```diff
@@ -14,109 +10,112 @@ var router = express.Router();
...
-                                                testdata: testdata});
+    // Render the page for running a test on a single component (i.e. continue to run this test on the same component)
+    res.render('test_run_singleComponent.pug', {testdata: test,
+                                                formId,
+                                                form,
+                                                componentUuid: test.componentUuid});
   }
   catch(err)
   {
-    logger.error(err);
```
micchickenburger, line 79: Sorry, I should have clarified!  Server-side errors write to `stderr`!  It’s browser/client-side console statements that should be removed.  If we need to log into a server to troubleshoot an issue, logging information is helpful.

routes/jobRoutes.js
```diff
@@ -1,155 +1,220 @@
...
+// Edit an existing job
+router.get('/job/:jobId([A-Fa-f0-9]{24})/edit', permissions.checkPermission("jobs:submit"), async function(req, res, next)
+{
+  try
+  {
+    // Retrieve the job's DB entry, using its job ID
+    // Throw an error if there is no DB entry corresponding to this job ID
+    var job = await Jobs.retrieve(req.params.jobId);
+    
+    if(!job)
```
micchickenburger, line 111: `if (!job) next();` is more succinct

static/css/dashboard.css
```diff
@@ -211,28 +211,16 @@ body.deployment-devsite .navbar-brand img {
...
-body.deployment-laptop {
-  background-image: linear-gradient(45deg, #f7f7d7 2.83%, rgba(255, 255, 255, 0) 2.83%, rgba(255, 255, 255, 0) 50%, #f7f7d7 50%, #f7f7d7 52.83%, rgba(255, 255, 255, 0) 52.83%, rgba(255, 255, 255, 0) 100%);   
-  background-size: 100vw 100vh;
-  background-repeat: no-repeat;
-}
-
 body.deployment-laptop #navbar-brand
 {
-  background-color: #f7f7d7 !important;
+  background-color: #fff3cd !important;
```
micchickenburger, line 216: Oh boy.  `!important` declarations in CSS are a major headache and slow page render performance.  We will have to tackle these at some point.

micchickenburger (dismissed):
Overall all looks fine.  Just nitpicks.  There is a lot of C-style here, which is atypical in the JavaScript/webdev world, but we can auto-address later when we get a linter in place.  Lots of `pre` statements, but you already know how I feel about those.  😁

pug/job.pug
krishmaj: Yes, `!{}` is pug string interpolation.  I'll add comments at all instances of it to describe.

pug/job_submit.pug
krishmaj: As above.

routes/draftRoutes.js
krishmaj: So then I shouldn't get rid of the `logger` statements?  Only the `console` and `debugger` ones?

static/css/dashboard.css
krishmaj: Sure.  They were already present, so I just left them as they are.  I'll link this comment to the CSS related issue as well.

routes/jobRoutes.js
krishmaj: Mmm, personally I still prefer even single-line statements to be indented in brackets, same as multi-line blocks.  Feel like it keeps things more consistent.  Chalk it up to stylistic differences?

pug/edit_jobTypeForm.pug
krishmaj: I think that's basically how it was originally actually, but I found that way very hard to read myself ... personally can't stand opening brackets at one indent level on a non-empty line, and closing them at a different level on an empty one (just a pet peeve really, but a strong one!).  I'll play around and perhaps find a neater looking compromise.

pug/job.pug
krishmaj: So just to be clear, something like this?

```
window.addEventListener('load', populateForm);

async function populateForm()
{
  ...
};
```

pug/job_submit.pug
krishmaj: For the `jobForm.setAlert("danger", statusMsg + " (" + statusCode + ")");` line you mean?

pug/job.pug
micchickenburger: Yup!

pug/job.pug
micchickenburger: Okay, if it’s a normal pug thing then maybe comments won’t be necessary.  :)

routes/draftRoutes.js
micchickenburger: Yeah :)

pug/edit_jobTypeForm.pug
krishmaj: Made it more Javascript-like now.

pug/job.pug
krishmaj: Done.

pug/job_submit.pug
krishmaj: Removed second condition, so it's now just `if(res.responseText)`.

pug/list_jobTypes.pug
krishmaj: Changed.

routes/draftRoutes.js
krishmaj: I've added the logger error messages back in where appropriate.  Note that I've done so only in `routes/jobRoutes.js` ... I've actually already removed this specific file (`routes/draftRoutes.js`) as part of the next commit I'm going to make.

krishmaj:
@micchickenburger I tried doing the Github identities stuff again yesterday, but it apparently hasn't made any difference.  I've sent you an email about that, but it looks like there's issues with my email service right now ... delays on sending and receiving.  (I only just got notifications a few minutes ago about your comments on here from yesterday!).  So who knows when you'll get it.

PR #12, opened 2022-01-03
Overhaul of the Test entity code:

krishmaj:
- cleaned up the backend library and routes code (partially addressing issue #10) - removed unused functions and code, console and logger statements apart from error handling, and added better comments
- cleaned up and improved layouts on frontend pug pages
- drafts are now exclusive to Tests, so combined draft related code (routes and list displays) into Test related equivalent files


krishmaj:
@micchickenburger I'm still not seeing a "Verified" next to my commit ... the heck's going on?!

krishmaj:
@micchickenburger Success!  That -S flag was the problem all along ... never had to sign my commits before, so I never even realised it was a thing!

lib/Tests.js
```diff
@@ -255,35 +209,42 @@ TestsModule.prototype.getRecentComponentTests = async function(componentUuid)
...
+  logger.info(query,"QfindTestIdStartsWith");
+ 
+  var matches = await db.collection("tests")
+    .find(query)  // binary ObjectIds only
+    .project({"_id":1, 'componentUuid':1, 'formId':1})
+    .limit(max)
+    .toArray();
+  // if(matches.length>=max) return new Error("Too many entries"); // too many!
+  for(var m of matches) {
+    try {  m.componentUuid = MUUID.from(m.componentUuid).toString(); } catch {};
```
micchickenburger, line 241: If we expect errors to occur here, why do we let them go silently?  We should either handle the errors or document why we don’t handle them.

lib/Tests.js
```diff
@@ -341,41 +302,113 @@ TestsModule.prototype.search = async function(txt,match,limit,skip)
...
+    record.state = record.state || 'submitted'; // ensure 'state' is set.
+    // logger.info("Tests::save()",record.formId,record.state);
+
+    if(!permissions.hasPermission(req,this._collection+':submit'))  throw new Error("You don't have permission to edit forms.");
+    
+    var result;
+    var draft_id = input._id;
+
+    // If it's a draft and already has a record number, replace it.
+    if(input.state == "draft" && draft_id) {
```
micchickenburger: Just a nitpick, but we should always use type-constrained comparisons:  `===`.

pug/list_testTypes.pug
```diff
@@ -1,123 +1,132 @@
...
                 span(aria-hidden = true) &times;
             .modal-body
               div
                 | Enter the new test type's form ID.
                 br
-                | NOTE: this ID should have only underscores and alphanumerics - the human-readable name will be set on the next page.
-                input.NewTestTypeName.form-control(type = "text")
+                | NOTE: this ID should have only underscores and alphanumerics (no spaces) - the human-readable name will be set on the next page.
+              pre #{''}
+              input.NewTestTypeName.form-control(type = "text")
```
micchickenburger, line 33: Just as an FYI:  we can add a `pattern` attribute to input elements for client-side input validation.  We can also automatically remove unwanted characters, ie:  `input.value.replace(/\s/g, '');` to replace all spaces with empty string.

pug/list_testTypes.pug
```diff
@@ -1,123 +1,132 @@
...
-              button.btn.btn-secondary(type = "button", data-dismiss = "modal") Close
-              button.btn.btn-primary.NewTestTypeButton(type = "button") Create
+              button.btn.btn-danger(type = "button", data-dismiss = "modal") Close
+              button.btn.btn-success.NewTestTypeButton(type = "button") Create
+      
       script.
         $('button.NewTestTypeButton').click(function()
         {
-          window.location.href = "/tests/" + $('button.NewTestTypeButton').parents('.modal-content').find('input').val() + "/new";
+          window.location.href = '/tests/' + $('button.NewTestTypeButton').parents('.modal-content').find('input').val() + '/new';
```
micchickenburger, line 41: This is curious… most modern web frameworks have client-side routing, but not pug?

pug/test.pug
```diff
@@ -49,135 +38,103 @@ block extrascripts
...
-          a(href = '?version=' + v)=v
-          |
-          |
-
-    pre #{''}
-    
-    if(processes && processes.length > 0)
-      h4 Associated Processes
+        
+        if(form.formName == "Wire Tensions")
```
micchickenburger: `===`

micchickenburger (dismissed):
Just some optional nits and questions.

lib/Tests.js
krishmaj: I'll address this when I clean up the 'autocomplete' route which uses this function.  But yes, you're absolutely correct of course - shouldn't ever have silent errors.

lib/Tests.js
krishmaj: Changed.  I'll clean up that 'save' function as a whole with the rest of the API-related functions sometime later.

pug/list_testTypes.pug
krishmaj: [sent you an email about this, since I'd like to close this PR and move onto the next one]

pug/test.pug
krishmaj: Changed.

PR #13, opened 2022-01-28
Overhaul of the Users related code:

krishmaj:
- removed the ability to edit user information in the DB interface ... this was a major security flaw, in that any user could in fact edit any other user's information - not just their own!  And these sorts of changes should be done through Auth0 anyway, not the DB
- removed the ability to promote users through the DB interface ... changes to user roles and permissions should be done centrally through Auth0 from now on so that we don't end up with the mess of roles and permissions that we currently have
- moved the current user's roles and permissions information to the Profile page ... reduces the amount of clutter on the homepage
- cleaned up and made formatting and layout consistent on the user and m2m user list pages


krishmaj:
Whoops - sorry, just saw that ```list_m2mUsers.pug``` is still using my old really-not-Javascript-like formatting ... will fix and push.

micchickenburger (approved):
All looks good to me.  Nice work 👍

PR #15, opened 2022-02-02
Changes relating to Component Short UUIDs and History:

### LINKED ISSUE #7: Component History is broken
Attempting to access the component history (from any individual component information page) gives an error.

krishmaj:
- components from now on will have their short UUID stored in their DB entry upon initial creation (unfortunately not retroactive, but doesn't really matter)
- component QR codes now encode the shortened URL, and this now redirects to the full URL of the component's main information page
- component short UUID is displayed on the information page if available, and if not available it is made explicitly clear
- the short UUID will only be used within QR codes, not as a redirect for every component related page (these additional routes will be phased out soon)
- cleaned up main component information page, as well as directly related component and test routes and libraries (rest to be overhauled soon)
- fixed component history ... didn't actually fix the original bug, but found a different way of showing previous component versions (this solves issue #7)


krishmaj:
Forgot to mention one thing in the original comment: the component history (list of previous versions) is now included in the main information page, so there is no need for a separate page and associated route.

micchickenburger (approved):
No major issues that I can find

PR #16, opened 2022-02-07
Overhaul of the Component entity code:

krishmaj:
- changed the 'component.type' field to 'component.formId', to be consistent with tests and jobs, and added 'component.formName' for same
- cleaned up non-API related 'Components' library functions, and made code structure consistent with 'Jobs' and 'Tests' libraries (also required slight modification to how the library is then called from other files)
- split the current user's recently visited components to a separate page, accessible from the sidebar (rather than busying up the homepage)
- cleaned up remaining component routes
- cleaned up component related pug files, and made formatting, layout and interface consistent with tests and jobs pages
- removed 'component_without_form' pug page (corresponding route was already removed) - we should NOT have the ability to create a component without a type
- removed 'contact_sheet' pug and associated route - as with above, we should NOT have the ablity to just dump a load of new components into the DB with no info apart from the UUID
- renamed component 'label' and 'traveller' pages to 'qrCodes' and 'summary' respectively ... more indicative of what they actually are ... and moved the previously separated 'traveller' .js layout into the actual .pug to be consistent with other similar pages


pug/edit_component.pug
```diff
@@ -150,191 +147,116 @@ block extrascripts
...
+    // Search the DB for any component of the same type as this submission that was most recently modified (created or edited) by this user
     function searchForRecentSubmission()
     {
       $('#loadRecentData').hide();
       
-      $.post("/json/search/component/" + (component.type || "") + "?limit=1",
-             {
-              "insertion.user.user_id": "!{user.user_id}",
-             })
+      $.post('/json/search/component/' + (component.formId || "") + "?limit=1", {"insertion.user.user_id": "!{user.user_id}"})
```
krishmaj, line 170: There is a bug here relating to the field name change (`component.type` -> `component.formId`).  For some reason, it is not recognising the new field name for components that definitely do have it, and therefore the button to load the recent data from such components is not appearing.  The wierd thing is that the button DOES appear for components with the old field name.  This is as far as I could trace it ... can't seem to find what happens after this point.

micchickenburger (approved):

PR #20, opened 2022-03-30
Various Changes

krishmaj:
CRITICAL bug fixes for user permissions ... without this the interface is completely broken
Additional changes to clean up some of the code, move some files around and remove old and unused files.


app/app.js
```diff
@@ -102,13 +102,13 @@ async function createApp(app) {
...
   app.use('/css',compileSass({
-      root: `${__dirname}/scss`,
+      root: `${__dirname}/static/scss`,
       sourceMap: true, // Includes Base64 encoded source maps in output css
       sourceComments: true, // Includes source comments in output css
       watchFiles: true, // Watches sass files and updates mtime on main files for each change
       logToConsole: false // If true, will log to logger.error on errors
   }));
-  app.use('/css',express.static('../scss'));
+  app.use('/css',express.static('../static/scss'));
```
micchickenburger: I know we talked about this via email.  It doesn’t matter so much where SCSS files are located; however, they are _not_ static resources.  They have to be compiled.

app/lib/permissions.js
```diff
@@ -65,7 +65,7 @@ function checkPermissionOrUserId(required_permission) {
     if(req.params.user_id == req.user.user_id) return next();
     if(hasPermission(req,required_permission)) return next();
     var p = required_permission;
-     return res.status(400).render('permissionsError.pug',{required_permission:p,required_user:req.params.user_id});
+     return res.status(400).render('user_permissionError.pug',{required_permission:p,required_user:req.params.user_id});
```
micchickenburger, line 68: Nitpick:  mix between camel and pascal case.  JavaScript world typically standardizes on camel case.

app/pug/user_profile.pug
```diff
@@ -52,12 +50,11 @@ block content
...
       dd
-        pre #{JSON.stringify(user.roles)}
-    
+        code #{JSON.stringify(user.roles)}
+      
       dt Permissions granted by these Roles:
       dd
-        <pre>#{JSON.stringify(user.permissions)}</pre>
-    
+        code #{JSON.stringify(user.permissions)}
```
micchickenburger: Semantic nitpick:  `code` is for displaying computer code.  If the idea is to just use a monospace font, use CSS.  If the semantic is preformatted text, `pre` is the correct element to use.  See here:  https://developer.mozilla.org/en-US/docs/Web/HTML/Element/code

micchickenburger (changes requested):
Two things are just nitpicks, but I do think the scss directory choice should be addressed.

app/app.js
krishmaj: Good point, yes - I didn't consider that.  Guess there isn't really any existing place where the scss files would fit, so I'll revert it back to how it was originally.

app/lib/permissions.js
krishmaj: Noted - will deal with that in the future ... there's quite a few other pug files with the same mix.  Personally I like that it makes it clear the broad ("user") vs. specific ("permissionError") purpose of the file, but that's just me I suppose.

app/pug/user_profile.pug
krishmaj: Noted, thanks.  Will change it to use CSS and add to the PR.

app/static/css/main.css
```diff
@@ -6,6 +6,10 @@ pre {
   white-space: pre-wrap;
 }
```
micchickenburger, line 8: I think you need a dot in here to specify a class (we can’t invent new HTML elements.)  Also, a better name for the class might just be `.monospace`

app/static/css/main.css
```diff
@@ -6,6 +6,10 @@ pre {
...
 
+codelike {
+  font-family: monospace, monospace;
```
micchickenburger: Listing monospace twice is redundant

app/pug/user_profile.pug
```diff
@@ -52,12 +50,11 @@ block content
...
       dd
-        pre #{JSON.stringify(user.roles)}
-    
+        codelike #{JSON.stringify(user.roles)}
+      
       dt Permissions granted by these Roles:
       dd
-        <pre>#{JSON.stringify(user.permissions)}</pre>
-    
+        codelike #{JSON.stringify(user.permissions).replace(/,/g, ", ")}
```
micchickenburger: As mentioned below we can’t invent new HTML elements.  Adding a dot in front would establish this as a class

micchickenburger (changes requested):

app/static/css/main.css
krishmaj: See here: https://stackoverflow.com/questions/38781089/font-family-monospace-monospace .

app/static/css/main.css
krishmaj: Changed in new commit.

app/pug/user_profile.pug
krishmaj: Changed in new commit.

app/static/css/main.css
micchickenburger: Ah, very interesting.  I think you should add a comment with that link to explain that since it’s not obvious and maybe not well-known.

krishmaj: Added.

micchickenburger (dismissed):

app/lib/constants.js
```diff
@@ -1,39 +1,43 @@
...
+  DB_CERTKEY: process.env.DATABASE_CERTKEY,    // Certificate & key in PEM format
+  
   // App Constants
   APP_PORT,
   NODE_ENV: process.env.NODE_ENV || 'development',
   BASE_URL: process.env.BASE_URL || `http://localhost:${APP_PORT}`,
   SESSION_SECRET: process.env.SESSION_SECRET || crypto.randomBytes(128).toString('hex'),
   M2M_SECRET: process.env.M2M_SECRET || crypto.randomBytes(256).toString('hex'),
-
+
```
micchickenburger: You’ve introduced a lot of rogue whitespace in this file.  My assumption is that you’re not using VS Code yet?  I wrote this file to be compatible with future styling configuration so it really should not be changing at all unless you’ve found a bug or are adding a new feature.

app/lib/email.js
```diff
@@ -1,28 +1,18 @@
...
-    ...arg
-  };
-  logger.info("sending email",data);
+module.exports = async function(arg)
+{
+  var data = {from   : EMAIL.FROM,
+              subject: "DUNE APA DB Message " + moment().toISOString(),
+              to     : EMAIL.ADMIN_RECIPIENTS,
+              ...arg};
+
```
micchickenburger: Also here you are moving from a standard JavaScript style to an nonstandard one.  Note also that when we implement linting all this work would be automated.

micchickenburger (changes requested):
I feel compelled to reject the changes in this recent commit since they are stylistic in nature (they don’t fix any bugs or add new features) and are contrary to typical style standards.  Implementing linting can help us automate all of these stylistic and syntactical changes to industry standards.

krishmaj:
As per our email discussion, I've completely reverted all the library files that I made only stylistic changes to, leaving only `Components,js`, `Tests.js` and `Jobs.js` using my non-standard styling, since I don't want to revert my bug fixes in those.  Very sorry about all that ... as I mentioned - I'm having trouble getting VS Code working on my computer (is a problem with the computer, not VS Code), so I haven't switched over to that yet.

micchickenburger (approved):

Issue #9, opened 2021-12-01
Replace non-CSS style hacks with proper CSS


krishmaj:

micchickenburger:
Reference pull request #8 review discussion:  https://github.com/DUNE/dunedb/pull/8#discussion_r760480075_

krishmaj:
@micchickenburger Would replacing the `pre #{''}` elements with appropriate numbers of `br` elements be satisfactory?  And was there anything else that you noticed needed replacing?

micchickenburger:
`br` elements are also hacks for using HTML to affect presentation.  The appropriate solution is CSS.

What do all of the elements have in common that extra space is needed?  Do they all have the same element after them?  If so, you can target that element in CSS and add a top margin.  Or you could add a bottom margin to their parent element.

For example:
```html
<main>
  <article>
    <!-- content -->
    <!-- need space here -->
  </article>
  <section>
    <!-- content -->
  </section>
</main>
```

```css
/* Option 1 */
section { margin-top: 1rem; }

/* Option 2 */
article { margin-bottom: 1rem; }
```

krishmaj:
There's no particular pattern or commonality as to where the extra space is needed - I just need to be able to drop in blank lines or some vertical spacing wherever it needs to be to improve the layout of the pages.  I'll keep exploring the options.

krishmaj:
I think it's worth pointing out that `br` and/or `pre` are in fact accepted ways of generating vertical whitespace within pug/Jade templates (which is really what we're doing here), simply because there is no built in way to otherwise do so.  So I don't necessarily agree with the assertion that it's wrong to use them in this specific circumstance.

That said, using `div(style = "margin-bottom: 0.Xcm;")` with an appropriate number in place of `X` looks to solve the issue using your preferred CSS styling?

micchickenburger:
Krish, they are not acceptable ways of adding whitespace, but people do it anyway.  Same goes for adding CSS directly to an HTML element like in your example.

It's all about maintainability.  One day when someone is trying to figure out why there is extra spacing somewhere they will burn time trying to find these things, and will have little confidence in making corrections.  The user experience will also lose consistency, which is really important.

In general there are several big "no-noes" when it comes to writing maintainable markup:
- Style-affecting attributes (e.g. `bgcolor`) should never be used
- The style attribute (e.g. `<article style="...">`) should only be used for dynamic style changes (e.g. animating an element with JavaScript for old browsers) and therefore rarely used
[...]

Issue #6, opened 2021-11-16
Non-default page titles not showing


krishmaj:
If a pug page has a page title variable set, it does not show in the browser tab - instead it displays the page address.  Does not happen if no variable is present - in that case, it shows the default "Sietch" title in the tab.

PR #21, opened 2022-04-04
Getting pug files up to coding standards ...

krishmaj:
- replaced non-CSS styling hacks with correct CSS styling - this should resolve issue #9
- moved previously embedded JavaScript code into new `/static/pages/` files
- moved previously embedded CSS styling into new `/static/css/` files
- cleaned up unnecessary whitespace and other formatting in pug files
- made code for determining page titles consistent for component, test and job list pages

Additional changes after 1st commit to resolve the following:
- removed `!important` CSS rule declarations which were being used in bad practice, and weren't actually doing anything anyway
- fixed bug with page titles not being shown in browser ... resolves issue #6


app/static/css/dashboard.css
```diff
@@ -1,225 +1,182 @@
...
+    -o-transform     : none !important;
+    -moz-transform   : none !important;
+    -webkit-transform: none !important;
+    transform        : none !important;
+    
+    -webkit-animation: none !important;
+    -moz-animation   : none !important;
+    -o-animation     : none !important;
+    animation        : none !important;
+}
```
krishmaj, line 22: @micchickenburger I remember a while ago you brought up something about having these `!important` tags here ... can you remind me what that was about, and if it's something that we can resolve as part of this PR?

krishmaj: Okay, so it actually looks like none of these `!important` tags are even needed - there are no other more specific rules in any of our CSS that relate to this `.disable-animations` class or any of its elements ... so nothing for the `!important` to even override.  Removing all of them makes no difference to the homepage (which is the only place where this class is used).

Will see about the other uses of this tag further down in the file.

app/static/css/dashboard.css
```diff
@@ -1,225 +1,182 @@
...
-
-.navbar-brand {
-  padding-top: .75rem;
-  padding-bottom: .75rem;
-  font-size: 1rem;
-  background-color: rgba(0, 0, 0, .25);
-  box-shadow: inset -1px 0 0 rgba(0, 0, 0, .25);
+body.deployment-development #navbar-brand {
+    background-color: #fff3cd !important;
+    color           : black;
```
krishmaj: Same with these ones - they don't actually make any difference since there are no other more specific rules for these style elements that would be overridden.

app/pug/common.pug
```diff
@@ -1,49 +1,17 @@
 
+block pugscripts
+  - function isAuthenticated()     { return !!(user); }
+  - function user_email(aUser)     { if (isAuthenticated()) return ((aUser.emails || [])[0] || {}).value || ""; else return "[Please login to see!]" };
+  - function user_shortname(aUser) { if (isAuthenticated()) return aUser.displayName || user_email(aUser); else return "[Please login to see!]" };
```
micchickenburger: `else` is unneeded on lines 4 and 5.

app/pug/common.pug
```diff
@@ -1,49 +1,17 @@
 
+block pugscripts
+  - function isAuthenticated()     { return !!(user); }
```
micchickenburger: This could be abbreviated:  `const isAuthenticated = () => !!user;`

app/pug/common.pug
```diff
@@ -1,49 +1,17 @@
...
+  - function user_email(aUser)     { if (isAuthenticated()) return ((aUser.emails || [])[0] || {}).value || ""; else return "[Please login to see!]" };
```
micchickenburger: I think this line is confusing and should have a comment describing what’s happening.

app/pug/splash.pug
```diff
@@ -1,17 +1,15 @@
...
   title DUNE APA Construction DB
-  link(rel = "preconnect" href = "https://fonts.gstatic.com")
-  link(href = "https://fonts.googleapis.com/css2?family=Inconsolata&family=Palanquin&display=swap", rel = "stylesheet")
-  link(rel = 'stylesheet' href = '/dist/formiojs/formio.full.min.css')
+  
+  link(rel = 'stylesheet', href = '/css/splash.css')
+  link(rel = 'stylesheet', href = '/dist/formiojs/formio.full.min.css')
+  link(rel = 'stylesheet', href = '/dist/bootstrap/css/bootstrap.min.css')
+  link(rel = 'stylesheet', href = 'https://fonts.googleapis.com/css2?family=Inconsolata&family=Palanquin&display=swap')
+  link(rel = 'preconnect', href = 'https://fonts.gstatic.com')
```
micchickenburger: Question:  Is there benefit to this prefetch if the fonts are being loaded from a different domain?

app/static/css/component.css
```diff
@@ -0,0 +1,14 @@
...
+.copybutton {
+    background-color: #008CBA;
+    width           : 45px;
+    text-align      : center;
+    color           : white;
+    font-size       : 10px;
+    padding         : 3px 3px;
+    border-radius   : 4px;
+    border          : 2px solid #008CBA;
+}
```
micchickenburger: File needs a newline at the end

app/static/css/component_qrCodes.css
```diff
@@ -0,0 +1,155 @@
...
+
+@media print {
+    .noprint {
+        visibility: hidden;
+    }
+    
+    .qr-row {
+        page-break-inside: avoid;
+    }
+}
```
micchickenburger: Missing newline at end of file

app/static/css/component_summary.css
```diff
@@ -0,0 +1,35 @@
...
+    transform-origin: top left;
+    transform       : rotate(90deg);}
+
+.section-content {
+    
+}
+
+.section-header {
+    margin-left: 44px;
+}
```
micchickenburger: Missing newline at end of file

app/static/css/dashboard.css
```diff
@@ -1,225 +1,182 @@
...
-  color: #333;
+    font-weight: 500;
+    color      : #333;
 }
 
 #sidebar i.fa {
-  font-size: .75rem;
-  margin-right: 3px;
-  color: #999;
+    font-size   : .75rem;
```
micchickenburger: This whole file:  Two spaces is de facto; colons should have no spaces to their left.

Proper:
```css
#sidebar .nav-link {
  font-size: 0.75rem;
}
```

app/static/css/main.css
```diff
@@ -45,19 +61,13 @@ pre {
...
-.border-editForm {
-    border: 3px solid red;
-    border-radius: 10px;
-    padding: 10px;
-}
+.formbuilder_area {
+    border       : 3px solid red;
+    border-radius: 10px; 
+    padding      : 10px;
+}
```
micchickenburger: Needs newline

app/static/pages/component_summary.js
```diff
@@ -0,0 +1,42 @@
...
+    var builtform  = $('.builtform', this);
+    var recordType = builtform.data("recordtype");
+    var record     = builtform.data("record");
+    
+    // Get the type form corresponding to the record
+    var schema_url = '';
+    
+    if(recordType === "component") { schema_url = "/json/componentForms/" + record.formId; }
+    else if(recordType === "test") { schema_url = "/json/testForms/" + record.formId; }
+    else if(recordType === "job")  { schema_url = "/json/jobForms/" + record.formId; }
```
micchickenburger: Prefer backtick notation:
```schema_url = `/json/componentForms/${record.formId}`;```

File: app/static/pages/edit_component.js [hunk 13]
```diff
@@ -0,0 +1,207 @@
...
+  
+  return obj;
+}
+
+// Search the DB for any component of the same type as this submission that was most recently modified (created or edited) by this user
+function searchForRecentSubmission()
+{
+  $('#loadRecentData').hide();
+  
+  $.post('/json/search/component/' + (component.formId || "") + "?limit=1", {"insertion.user.user_id": "!{user.user_id}"})
```
micchickenburger: Prefer backtick notation for readability

micchickenburger (changes requested):
In all the CSS files it looks like you inadvertently changed the spacing from two to four spaces.  Also, colons should abut their property names.  Finally, newlines should exist in all files.

app/pug/common.pug
krishmaj: So by that you mean the syntax is `if (condition) return [value1]; return [value2]`?

app/pug/common.pug
krishmaj: Is that some kind of JavaScript specific syntax?  Never seen something like that before.

app/pug/splash.pug
krishmaj: Honestly have no idea ... it was already like that, so I just left it as is, apart from cleaning up the code itself.  I can dive a bit deeper into it and get back to you.

app/static/css/dashboard.css
krishmaj: Sorry about that ... I think that the two-space thing (and the missing newlines) may have happened because I was editing the files using two different text editors, which have different space/tab and end-of-file settings.  Seen it happen before for the same reason.  Will fix.

app/static/pages/component_summary.js
krishmaj: Should I change it everywhere in the code right now?  Pretty sure there's a lot more places where I've used `"" +` type notation ... everywhere where there's an `href` for example.

Or I can change it here specifically right now, and make a new issue to do the same changes across the rest of the code.

app/pug/common.pug
krishmaj: I'll actually simplify the return statement to: `return (user.emails[0].value);` .  The current syntax assumes that a user might not have an associated email address, which used to be true with the old Sietch-related Auth0 tenant (never enforced requiring email addresses).  But with the new Auth0 setup this is never the case any more (since we now require users to have at least one).  Will still add a comment to describe the function anyway.

app/pug/common.pug
micchickenburger: Yup!

krishmaj: Huh ... even after 6 months JavaScript continues to feel weird to me!

app/pug/common.pug
micchickenburger: They are called arrow functions and differ from regular functions only in local scope.

```javascript
// Traditional syntax
function isAthenticated() {}

// Function assignment
const isAuthenticated = function () {};

// Arrow function
const isAuthenticated = () => {};

// Arrow functions with one statement need no curly braces,
// and the return value of the statement becomes
// the return value of the function
const area = (radius) => Math.PI * (radius * radius);

// Functions can also be anonymous
(function () {})(); // traditional
(()=>{})(); // arrow
```

app/static/pages/component_summary.js
micchickenburger: Just for new changes is fine for now I think.  We’ll automate these changes when we introduce linting.

app/pug/splash.pug
krishmaj: Looks like the `preconnect` line isn't needed, no ... as you say, the font stylesheet is already being loaded in the previous line anyway, and it doesn't appear to make any difference when removed.

There's also an identical situation in `dashboard.pug` - same two lines, with the `preconnect` one apparently making no difference at all.  I'll remove that one as well.

micchickenburger: Preconnect statements enable browsers to establish connections to external domains before reading statements to load resources.  It can improve UX.  https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/preconnect

krishmaj: Right, I did read that page ... didn't fully understand the importance I guess.  So pointing the `preconnect` lines to `fonts.gstatic` doesn't do anything since that's not where the fonts are being loaded from ... they should point to `fonts.googleapis`?  And the `preconnect` line should go before that one, not after?

krishmaj: Looking at it some more, would the `preload` link type be more appropriate here, instead of `preconnect`?  The MDN page for the former specifically uses the example of loading fonts from a CSS as an example of how using it can give performance improvement.

app/pug/common.pug
micchickenburger: The same applies in C too.  The statement after the condition executes in all cases, except for the condition because there is a return statement inside of it.

app/pug/splash.pug
micchickenburger: Sure, preload makes more sense I think.

micchickenburger (approved):

PR #22, opened 2022-04-13
Removing documents and files from DB

krishmaj:


app/lib/database.js
```diff
@@ -192,19 +192,6 @@ async function initialize_database(db)
...
-	  	await db.collection("docs").createIndex({docId:1});
-	  	await db.collection("docs").createIndex({"validity.version":1});
-	  	await db.collection("docs").createIndex({"validity.startDate":1});
-		 	await db.collection("docs").createIndex({"data": "text"});
-	  	logger.info("Indexed docs collection")
-	   	status_obj.version =7.0;
-		}
-
 	} catch(err) {
 		console.error(err);
```
micchickenburger, line 196: Why use console when we have a logger statement below?

micchickenburger (approved):

app/lib/database.js
krishmaj: Didn't actually pay any attention to the rest of the file, only on removing the bit dealing with documents.  Pretty sure there's quite a few other places in the code where it's still mixing between logger and console ... figure we can get rid of the remaining latter all in one go at some point (is already an active issue #10).

PR #27, opened 2022-05-19
Critical Bug Fix for DB Entity Form Retrieval

krishmaj:
Retrieval is currently hidden behind an uneeded permission - removed permission requirement.  Will not affect other permissions, since the page where retrieval occurs has its own specific permission requirement.


krishmaj:
@micchickenburger I know this file is a complete mess - should be split into separate files for component, test, job, etc. API routes like the non-API routes are.  I'll add that to the list of things I'll do when I'm back working on the DB full-time, but for right now I'd just like to get this bug fix in, if that's okay.

micchickenburger (approved):

PR #30, opened 2022-05-27
Adding ability to create batch-type components ...

krishmaj:
- additions to the route for creating new components to pass through a pre-generated list of UUIDs
- overhauled the component editing page, to allow creation and submission of individual sub-components alongside a batch-type component (for non-batch-type components, nothing has changed)
- added checkbox to the list of available Formio form components
- added batch affirmation checkbox to the component type form template
- table of sub-components will now be displayed on the component information page of batch-type components, and the originating batch UUID will be displayed on components that were created via a batch
- moved the component type to be displayed first among the component properties on the information page


krishmaj:
Just tagging on an additional low priority change in the code as well - we currently have a directory `static/components` which contains code relating to the Formio form components ... I've renamed this to `static/formio`, since we already have plenty of things called 'components' in the code and DB in general.  The actual contents of the directory need cleaning up at some point (I think some of the Formio components are no longer used, so could be removed) but that is also low priority and can be done at a later time.

micchickenburger (approved):

PR #32, opened 2022-06-02
Beginning overhaul of courses ... now called workflows:

krishmaj:
- cleaned up the library file
- cleaned up the routes file and added missing basic functionality, mirroring components, test and jobs
- updated naming in other library and routes files


app/lib/database.js
```diff
@@ -183,12 +183,12 @@ async function initialize_database(db)
...
-	  	db.collection("courses").createIndex({"validity.startDate":1});
-	  	logger.info("Indexed courses collection")
+	  	await ensure_exists(db,"workflows")
+	  	db.collection("workflows").dropIndexes();
+	  	db.collection("workflows").createIndex({workflowId:1});
+	  	db.collection("workflows").createIndex({"validity.version":1});
+	  	db.collection("workflows").createIndex({"validity.startDate":1});
+	  	logger.info("Indexed workflows collection")
 	   	status_obj.version =6.0;
 		}
```
micchickenburger, line 193: Is the database versioned?  If so we may need to increment the version for this change.

micchickenburger (approved):
Increment database version if versioning is used.  Otherwise it looks good to me.

app/lib/database.js
krishmaj: No idea actually.  Never seen anywhere where it seemed to matter, or even where the version was displayed (maybe buried in the logger output somewhere).  I'll merge this PR for now, and if we find that the versioning is in fact still being used I'll increment later.

PR #36, opened 2022-06-10
Implementing Actions + couple of other small changes

krishmaj:
- new library, routes and pug files for actions
- added action history to component information page
- added action pages to interface sidebar
- added references where relevant to action and action type forms in existing Forms library
- QoL change - components from trashed component types will no longer show in the lists of components (equivalent functionality for actions is included in new files)
- re-implemented type form tags (previously also called categories) ... code and interface pages are fully functional, but sidebar link is disabled for the moment, until we know more about how people intend to use tags (if at all)
- cleaned up remaining miscellaneous routes, and renamed to start routes (since that's all that's left in that file now)


app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+    delete newRecord.insertion;
+  }
+
+  newRecord.insertion = commonSchema.insertion(req);
+
+  var _lock = await dbLock("saveAction" + newRecord.actionId, 1000);
+
+  // Attempt to retrieve an existing record with the same action ID as the specified one
+  // This is relevant if we are editing an existing record
+  // If no such record exists, i.e. we are saving a completely new action, this will return 'null'
```
micchickenburger, line 57: For the future, comments like this are unnecessary since the code says exactly the same thing.

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+  }
+
+  // Set up a new record using the provided 'input' object as a basis
+  var newRecord = { ...input };
+
+  // Add the type form name to the new record
+  // If no type form name has been specified in the 'input' object, use the one directly from the type form itself
+  if (!input.typeFormName) {
+    newRecord.typeFormName = typeForm.formName;
+  }
```
micchickenburger: This whole clause can be summarized like so:

```javascript
newRecord.typeFormName = newRecord.typeFormName || typeForm.formName;
```

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+    throw new Error("Actions::save() - you do not have permission [actions:perform] to create and/or edit actions!");
+  }
+
+  // Check that the minimum required information has been provided for a record to be saved
+  // For action records, these are:
+  //   - the action type form ID
+  //   - the UUID of the component on which the action has been performed
+  //   - user-provided data (may be empty of content, but must still exist)
+  if (!input) {
+    throw new Error("Actions::save() - the 'input' object has not been specified!");
```
micchickenburger: Technically this error is not correct.  The condition is actually testing for a falsey value, which `undefined` and `null` are both cast to.  But this statement would also execute for `false`, `0`, and other values.  Meanwhile an empty object `{}` or empty array `[]` would pass this test.

micchickenburger: image

micchickenburger: I'd recommend using `if (input instanceof Object)` instead.  And for properties below, `if (input.hasOwnProperty('typeFormId'))`.

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+const ObjectID = require('mongodb').ObjectID;
+const permissions = require('lib/permissions.js');
+
+module.exports = {
+  save,
+  retrieve,
+  versions,
+  list,
+  search,
+  autoCompleteId
```
micchickenburger: Prefer all last members of lists to also include a comma at the end.

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+
+const commonSchema = require('lib/commonSchema.js');
+const { db } = require('./db');
+const dbLock = require('lib/dbLock.js');
+const Forms = require('lib/Forms.js');
+const MUUID = require('uuid-mongodb');
+const ObjectID = require('mongodb').ObjectID;
+const permissions = require('lib/permissions.js');
+
+module.exports = {
```
micchickenburger: Prefer `module.exports` to exist at the end of files.  Prefer functions be defined before they are used.

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
+'use strict';
+
+const commonSchema = require('lib/commonSchema.js');
+const { db } = require('./db');
+const dbLock = require('lib/dbLock.js');
+const Forms = require('lib/Forms.js');
+const MUUID = require('uuid-mongodb');
+const ObjectID = require('mongodb').ObjectID;
```
micchickenburger: Prefer library imports be listed before file imports.

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+  if (!input.data) {
+    throw new Error("Actions::save() - the 'input.data' has not been specified!");
+  }
+
+  // Check that there is an existing type form corresponding to the the provided type form ID
+  var typeFormsList = await Forms.list('actionForms');
+  var typeForm = typeFormsList[input.typeFormId];
+
+  if (!typeForm) {
+    throw new Error("Actions:save() - the specified 'input.typeFormId' (" + input.typeFormId + ") does not match a known action type form!");
```
micchickenburger: Prefer backtick notation:  ``throw new Error(`Actions:save() - the specified 'input.typeFormid' (${input.typeFormId}) does not...`);``

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+  if (!input.componentUuid) {
+    throw new Error("Actions::save() - the 'input.componentUuid' has not been specified!");
+  }
+
+  if (!input.data) {
+    throw new Error("Actions::save() - the 'input.data' has not been specified!");
+  }
+
+  // Check that there is an existing type form corresponding to the the provided type form ID
+  var typeFormsList = await Forms.list('actionForms');
```
micchickenburger, line 27: Prefer to never use `var`.  Use `const` unless value will be changed, then use `let`.

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+  var typeForm = typeFormsList[input.typeFormId];
+
+  if (!typeForm) {
+    throw new Error("Actions:save() - the specified 'input.typeFormId' (" + input.typeFormId + ") does not match a known action type form!");
+  }
+
+  // Set up a new record using the provided 'input' object as a basis
+  var newRecord = { ...input };
```
micchickenburger, line 33: Consider whether starting with an empty object `{}` is a simpler, more maintainable implementation than the current implementation of destructuring an arbitrary input with potentially unsanitized members and deleting some set of members post hoc.

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+  // If no type form name has been specified in the 'input' object, use the one directly from the type form itself
+  if (!input.typeFormName) {
+    newRecord.typeFormName = typeForm.formName;
+  }
+
+  // Make sure that the new record's 'componentUuid' field is in the correct format (it may not be so in the 'input' object)
+  newRecord.componentUuid = MUUID.from(input.componentUuid);
+
+  // Add additional information to the new record
+  newRecord.actionId = (input.actionId) ? new ObjectID(input.actionId) : new ObjectID();
```
micchickenburger: This complicated ternary is equivalent to `newRecord.actionId = new ObjectId(input.actionId);`

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+  delete newRecord._id;
+
+  // Insert the new record into the 'actions' records collection
+  var result = await db.collection('actions')
+    .insertOne(newRecord);
+
+  _lock.release();
+
+  // Throw an error if the insertion fails
+  if (result.insertedCount != 1) {
```
micchickenburger: Prefer type-strict conditionals:  `!==`

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+  // Return the record as proof that it has been saved successfully
+  return result.ops[0];
+}
+
+
+/// Retrieve a single version of an action record (either the most recent, or a specified one)
+async function retrieve(actionId, projection) {
+  // Construct the 'match_condition' to be used as the database query
+  // For this function, it is that a record's action ID must match the specified one
+  var match_condition = { actionId: actionId };
```
micchickenburger: This line is equivalent to `const match_condition = { actionId };`.

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+
+  // If there is at least one matching record ...
+  if (records.length > 0) {
+    // Convert the 'componentUuid' of the first matching record from binary to string format, for better readability and consistent display
+    records[0].componentUuid = MUUID.from(records[0].componentUuid).toString();
+
+    // Return the first matching record
+    return records[0];
+  }
+  else {
```
micchickenburger: Unnecessary `else`

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+  }
+
+  // Next we want to remove all but the most recent version of each matching record
+  // First sort the matching records by validity ... highest version first
+  aggregation_stages.push({ $sort: { 'validity.insertDate': -1 } });
+
+  // Then group the records by whatever fields will be subsequently used
+  // For example, if the 'actionId' of each returned record is to be used later on, it must be one of the groups defined here
+  // '$first' indicates that we take only the first record when they are grouped according to that field (the order remains highest version first)
+  // '$last' indicates that we take only the last record when they are grouped according to that field
```
micchickenburger, line 174: These `$first` and `$last` comments describe well-known, documented functionality and are therefore unnecessary

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+  // Note that this changes some field access via dot notation - i.e. in the returned records, 'action.data.name' becomes 'action.name'
+  aggregation_stages.push({
+    $group: {
+      _id: { actionId: '$actionId' },
+      actionId: { '$first': '$actionId' },
+      componentUuid: { '$first': '$componentUuid' },
+      typeFormId: { '$first': '$typeFormId' },
+      typeFormName: { '$first': '$typeFormName' },
+      name: { '$first': '$data.name' },
+      insertDate: { '$last': '$insertion.insertDate' }
```
micchickenburger: Prefer comma at the end of last members in lists (this line and next)

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+  }
+
+  // Return the entire list of action records
+  return records;
+}
+
+
+/// Search for action records
+/// The search can be performed via either a text search or specifying a record to match to
+async function search(textSearch, matchRecord, skip, limit) {
```
micchickenburger: You can remove the next few lines by setting default parameter values:

```javascript
async function search(textSearch, matchRecord, skip = 0, limit = 20) {
```

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+  // Otherwise, it is that a record must match the specified one
+  var match_condition = { ...matchRecord } || {};
+
+  // If the condition contains a component UUID, set it to binary format
+  if (match_condition.componentUuid) {
+    match_condition.componentUuid = MUUID.from(match_condition.componentUuid)
+  }
+
+  // If we are doing a text search, set the 'text' field of the condition
+  if (textSearch) {
```
micchickenburger: A lot of these short conditions can be moved to one line:

```javascript
if (textSearch) matchCondition['$text'] = { $search: textSearch };
```

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+
+  // Convert the 'componentUuid' of each matching record from binary to string format, for better readability
+  // Additionally, add a 'route' field to each record, which is the URL of that record's information page
+  // Make these changes on copies of the retrieved records, so that the records themselves are not changed
+  var searchResults = [];
+
+  for (var record of records) {
+    var result = { ...record };
+    result.componentUuid = MUUID.from(record.componentUuid).toString();
+    result.route = '/action/' + result.actionId.toString();
```
micchickenburger, line 250: Prefer backtick notation over string concatenation.

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+  // Return the finalised search results
+  return searchResults;
+}
+
+
+/// Auto-complete an action ID string as it is being typed
+/// This actually returns the records of all actions with a matching action ID to that being typed
+async function autoCompleteId(inputString, typeFormId, limit) {
+  // Set the maximum number of matching action IDs to display
+  limit = limit || 10;
```
micchickenburger: Can set default parameter value in function declaration

app/lib/Actions.js
```diff
@@ -0,0 +1,368 @@
...
+  var bithigh = ObjectID(q.padEnd(24, 'F'));
+
+  // Construct a 'match_condition' to be used as the database query
+  // For this function, it is that the action ID's binary value is between the minimum and maximum binary values defined above
+  var match_condition = {
+    actionId: {
+      $gte: bitlow,
+      $lte: bithigh
+    },
+    typeFormId: null
```
micchickenburger, line 277: Won't this only return results where `typeFormId` is explicitly set to `null` if it's not passed to the function?  Is that the intended functionality or should `typeFormId` not be relevant to the query at all?

app/pug/list_components.pug
```diff
@@ -29,20 +29,21 @@ block content
         
         tbody
           each component in components
-            - var uuid = MUUID.from(component.componentUuid).toString();
+            if((!singleType && !forms[component.formId].tags.includes("Trash")) || (singleType))
```
micchickenburger, line 32: The parenthesis around `singleType` are not necessary and confused me.  Consider breaking out into separate variable for readability:

```javascript
- const isNotSingleTypeOrInTrash = !singleType && !forms[component.formId].tags.includes('Trash');
- if (isNotSingleTypeOrInTrash || singleType)
```

You would know what name would be more appropriate.

app/pug/list_components.pug
```diff
@@ -29,20 +29,21 @@ block content
         
         tbody
           each component in components
-            - var uuid = MUUID.from(component.componentUuid).toString();
```
micchickenburger, line 32: Prefer to never use `var`

app/pug/list_jobTypes.pug
```diff
@@ -5,33 +5,42 @@ block vars
...
-              div
-                | Enter the new job type's form ID.
-                br
-                | NOTE: this ID should have only underscores and alphanumerics (no spaces) - the human-readable name will be set on the next page.
-              div.vert-space-x1
-                input.NewJobTypeName.form-control(type = "text")
-            .modal-footer
-              button.btn.btn-danger(type = "button", data-dismiss = "modal") Close
-              button.btn.btn-success.NewJobTypeButton(type = "button") Create
+      //- a(href = '#', data-toggle = "modal" data-target = "#NewJobTypeModal").NewJobType.btn.btn-warning
```
micchickenburger, line 21: Prefer removal of code over comments.  You can go back through git history to restore at a later time.

app/pug/list_testTypes.pug
```diff
@@ -5,34 +5,43 @@ block vars
...
+    //-         .modal-body
+    //-           div
+    //-             | Enter the new test type's form ID.
+    //-             br
+    //-             | NOTE: this ID should have only underscores and alphanumerics (no spaces) - the human-readable name will be set on the next page.
+    //-           div.vert-space-x1
+    //-             input.NewTestTypeName.form-control(type = "text")
+    //-         .modal-footer
+    //-           button.btn.btn-danger(type = "button", data-dismiss = "modal") Close
+    //-           button.btn.btn-success.NewTestTypeButton(type = "button") Create
```
micchickenburger, line 41: As above, prefer removal

app/routes/actions.js
```diff
@@ -0,0 +1,254 @@
...
+
+const Actions = require('lib/Actions.js');
+const Components = require('lib/Components.js')('component');
+const Forms = require('lib/Forms.js');
+const logger = require('../lib/logger');
+const permissions = require('lib/permissions.js');
+const router = require('express').Router();
+const utils = require('lib/utils.js');
+
+var default_form_schema = JSON.parse(require('fs').readFileSync('./schemas/default_form_schema.json'));
```
micchickenburger: Prefer never use `var`.  Prefer `require` statements at document head.

app/routes/actions.js
```diff
@@ -0,0 +1,254 @@
...
+
+
+/// View a single action record
+router.get('/action/:actionId([A-Fa-f0-9]{24})', permissions.checkPermission('actions:view'), async function (req, res, next) {
+  try {
+    // Set up a database query that includes the specified action ID and a version number if also provided
+    var query = { actionId: req.params.actionId };
+
+    if (req.query.version) {
+      query['validity.version'] = parseInt(req.query.version);
```
micchickenburger: Prefer including radix in `parseInt`:  `parseInt(req.query.version, 10);`

app/routes/actions.js
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    // Retrieve the specified version of the record using the query
+    // Simultaneously, retrieve ALL versions of the same record
+    let [action, actionVersions] = await Promise.all([
+      Actions.retrieve(query),
+      Actions.versions(req.params.actionId)
```
micchickenburger: Prefer comma at end of last list member

app/routes/actions.js
```diff
@@ -0,0 +1,254 @@
...
+    ]);
+
+    // Throw an error if there is no record corresponding to the query
+    if (!action) {
+      return res.status(400).render("There is no action record with action ID = " + req.params.actionId);
```
micchickenburger: Prefer backtick notation over string concatenation

app/routes/actions.js
```diff
@@ -0,0 +1,254 @@
...
+    // Render the interface page for viewing an action record
+    res.render('action.pug', {
+      action,
+      actionVersions,
+      actionTypeForm,
+      component
+    });
+  } catch (err) {
+    logger.error(err);
+    res.status(400).send(err.toString());
```
micchickenburger: Status code for server-side errors should be in the `500`s.

app/routes/actions.js
```diff
@@ -0,0 +1,254 @@
...
+    if (!action) {
+      return res.status(400).render("There is no action record with action ID = " + req.params.actionId);
+    }
+
+    // Retrieve the action type form, using its type form ID (which is specified in the record)
+    var actionTypeForm = await Forms.retrieve('actionForms', action.typeFormId);
+
+    // Throw an error if there is no type form corresponding to the type form ID
+    if (!actionTypeForm) {
+      return res.status(400).send("There is no action type form with form ID = " + action.typeFormId);
```
micchickenburger: `404` is I think a more appropriate response... It seems the request wasn't bad or malformed, the resource just wasn't found.

app/routes/api/index.js
```diff
@@ -4,6 +4,7 @@
 // Probably should be put into seperate files in routes/api/componentApi.js, etc
 // and loaded from this file.
 "use strict";
+const Actions = require('lib/Actions.js');
 const express = require("express");
```
micchickenburger, line 8: Prefer libraries be loaded before local files.  Prefer quotation consistency, preferably single quotes in JavaScript.

app/routes/components.js
```diff
@@ -1,4 +1,5 @@
 
+const Actions = require('lib/Actions.js');
 const Components = require('lib/Components.js')('component');
 const express = require('express');
```
micchickenburger, line 4: Prefer libraries be loaded before local files

app/routes/components.js
```diff
@@ -58,15 +59,19 @@ router.get('/component/' + utils.uuid_regex, permissions.checkPermission("compon
...
     //  - any related components
     // Throw an error if there is no DB entry corresponding to this form ID
-    let [form, tests, testForms, relatedComponents] = await Promise.all(
+    var match_condition = { componentUuid: req.params.uuid };
+
+    let [form, actions, actionForms, tests, testForms, relatedComponents] = await Promise.all(
     [
       Forms.retrieve("componentForms", formId),
+      Actions.list(match_condition),
+      Forms.list("actionForms"),
```
micchickenburger, line 72: Prefer single quotes in JavaScript

app/routes/components.js
```diff
@@ -93,6 +98,8 @@ router.get('/component/' + utils.uuid_regex, permissions.checkPermission("compon
     res.render("component.pug", {component,
```
micchickenburger, line 98: I know you're just filling in the blanks for this new feature, but a more readable notation could be:

```javascript
res.render('component.pug', { one, two, three, four, five, six, seven,
  eight, nine, ten, eleven, twelve });

// or

const componentList = {
  one,
  two,
  // ...
  eleven,
  twelve,
};
res.render('component.pug', componentList);
```

app/routes/start.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
```
micchickenburger: Note that this statement has no effect in modules:  http://www.ecma-international.org/ecma-262/6.0/#sec-strict-mode-code

app/routes/start.js
```diff
@@ -0,0 +1,27 @@
...
+
+const router = require('express').Router();
+
+module.exports = router;
+
+
+/// View the home page
+router.get('/', async function (req, res, next) {
+  // If the user is not logged in, go to the next function below
+  if (!req.user) {
```
micchickenburger: Can be abbreviated to `if (!req.user) return next();`

app/routes/start.js
```diff
@@ -0,0 +1,27 @@
...
+    return next();
+  }
+
+  // If the user is logged in and they have a specified start page, redirect there
+  if (req.user.user_metadata && req.user.user_metadata.start_page) {
```
micchickenburger, line 10: Is it possible for `req.user` to ever be undefined?  If so this will cause a runtime error.

app/routes/tags.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+const Forms = require('lib/Forms.js');
+const logger = require('../lib/logger');
+const router = require('express').Router();
```
micchickenburger: Prefer libraries be loaded before local resources

app/routes/tags.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
```
micchickenburger: Has no effect in modules

app/routes/tags.js
```diff
@@ -0,0 +1,32 @@
...
+
+const Forms = require('lib/Forms.js');
+const logger = require('../lib/logger');
+const router = require('express').Router();
+
+module.exports = router;
```
micchickenburger: Prefer module exports at end of file

app/routes/tags.js
```diff
@@ -0,0 +1,32 @@
...
+      Forms.list('componentForms'),
+      Forms.list('actionForms')
+    ]);
+
+    // Render the interface page for listing the tags which are associated with specific entity type forms
+    // The 'filtering' of the entity type forms with respect to individual tags is performed on the interface page itself
+    res.render('tags.pug', {
+      typeFormTags,
+      componentTypeForms,
+      actionTypeForms
```
micchickenburger: Prefer comma at end of last list member

app/routes/tags.js
```diff
@@ -0,0 +1,32 @@
...
+    });
+  } catch (err) {
+    logger.error(err);
+    res.status(400).send(err.toString());
```
micchickenburger: Any error caught here would be server-side and should return a `500` level error.

app/static/pages/action_editTypeForm.js
```diff
@@ -0,0 +1,245 @@
+// Set the flag for whether to use the built-in Formio form builder or not
+var builder_wizard = false;
+
+// Set up the 'metaschema' block using Formio form components
+// This block will appear at the top of the 'Edit Action Type Form' page
+var metaschema =
+{
```
micchickenburger: Prefer opening brace on same line as variable declaration, reducing indentation of entire object by two spaces

app/static/pages/action_editTypeForm.js
```diff
@@ -0,0 +1,245 @@
+// Set the flag for whether to use the built-in Formio form builder or not
+var builder_wizard = false;
+
+// Set up the 'metaschema' block using Formio form components
+// This block will appear at the top of the 'Edit Action Type Form' page
+var metaschema =
+{
+  "components": [
+    {
```
micchickenburger: Prefer opening curly brace to be adjacent to opening square brace: `[{`

app/static/pages/action_editTypeForm.js
```diff
@@ -0,0 +1,245 @@
...
+              "type": "textfield",
+              "label": "Type Form Name:",
+              "key": "formName",
+              "validate": {
+                "required": true,
+                "unique": false,
+                "multiple": false
+              },
+              "input": true,
+            }]
```
micchickenburger: Prefer comma at last list member

app/static/pages/action_editTypeForm.js
```diff
@@ -0,0 +1,245 @@
+// Set the flag for whether to use the built-in Formio form builder or not
+var builder_wizard = false;
+
+// Set up the 'metaschema' block using Formio form components
+// This block will appear at the top of the 'Edit Action Type Form' page
+var metaschema =
+{
+  "components": [
```
micchickenburger: This is not JSON.  Prefer no quotes for keys.  Prefer single quotes for string values.

app/static/pages/action_listTypes.js
```diff
@@ -0,0 +1,11 @@
+// When the user enters a new type form ID and presses the 'Create' button, redirect them to the page for creating the new type form
+$('button.NewTypeButton').click(function () {
+  window.location.href = '/actionTypes/' + $('button.NewTypeButton').parents('.modal-content').find('input').val() + '/new';
```
micchickenburger, line 3: This could break silently over time if any additional input element is added to `.modal-content`.  Better to be specific about the input value to extract, perhaps through a class name or ID attribute.  In that way you could also eliminate the use of jQuery here, which is always preferable.

app/static/pages/action_listTypes.js
```diff
@@ -0,0 +1,11 @@
...
+$('button.NewTypeButton').click(function () {
+  window.location.href = '/actionTypes/' + $('button.NewTypeButton').parents('.modal-content').find('input').val() + '/new';
+})
+
+// When the user enters a new type form ID and presses the 'Enter' key, redirect them to the page for creating the new type form
+$('input').on("keyup", function (e) {
+  if (e.key === "Enter") {
+    window.location.href = '/actionTypes/' + $(this).val() + '/new';
+  }
+})
```
micchickenburger: Missing semicolon.  Also, I don't see the advantage of using a library (jQuery) for this.  The native code looks almost identical and will be more performant:

```javascript
document.querySelector('input').addEventListener('keyup', function (e) {
  if (e.key === 'Enter') {
    window.location.href = `/actionTypes/${this.value}/new`;
  }
});
```

Also, have we verified that this is and will only ever be the only input element on the page?

app/static/pages/action_specComponent.js
```diff
@@ -0,0 +1,96 @@
+// Declare a variable to hold the completed type form that will eventually be submitted to the database
+var typeForm = null;
```
micchickenburger: In this implementation this is equivalent to `let typeForm;` and then appears as intentionally unassigned rather than intentionally assigned value `null`.  I searched for null checks but there weren't any, so this value doesn't seem to need to be set to anything at all.

app/static/pages/action_specComponent.js
```diff
@@ -0,0 +1,96 @@
...
+async function onPageLoad() {
+  // Set up a new type form based on the schema of the action type form
+  var schema = actionTypeForm.schema;
+
+  // Add a 'Submit' button to the type form (this is temporary, only in the scope of this script)
+  schema.components.push({
+    type: "button",
+    theme: "btn btn-success",
+    label: "Submit Action Record",
+    key: "submit",
```
micchickenburger: Prefer single quotes in JavaScript

app/static/pages/action_specComponent.js
```diff
@@ -0,0 +1,96 @@
...
+
+
+  /// Function to run for a failed submission
+  function postFail(result, statusCode, statusMsg) {
+    // If the submission result contains a response message, display it along with the appropriate Formio alert type
+    // Otherwise, display any status message and error code instead
+    if (result.responseText) {
+      typeForm.setAlert('danger', result.responseText);
+    } else {
+      typeForm.setAlert('danger', statusMsg + " (" + statusCode + ")");
```
micchickenburger, line 90: Prefer backtick notation

micchickenburger (approved):
I have recommended a number of code quality improvements but none of them are necessary for merge, assuming there aren't any bugs resulting from some of them.  Please review and use your discretion.  In particular, I noticed many of these issues (and did not mark all of them):

- In JavaScript, you are allowed to add a comma to the last element of a list if it is followed by a new line.  For example:
  ```javascript
  const example = [
    'one',
    'two',
    'three',
  ];
  ```
  The benefit of this is that if you need to add another list item in the future, git diff will show only the new line, not a new line plus a modified line.
- There is still extensive use of `var` but it is better to explicitly use `const` or `let` since these have explicit scope and mutability properties.
- `'use strict';` is often used in modules but has no effect.  Strict mode is always enabled for ECMAScript modules.
- Some objects have quotes around the keys, making them look like JSON.  This might be valid but it's confusing to read because it looks like a multiline string instead of a JavaScript object.
- There is inconsistent usage of single and double quotes. In the JavaScript world we prefer single quotes.
- There is large use of string concatenation.  I think backtick notation is more readable and eslint will agree.
- There is some unnecessary use of jQuery
- Some conditionals could be made shorter by removing the braces and moving to one line
- Some HTTP response codes I think should change
- Module exports are normally located at the bottom of the module and would ensure we aren't using a variable name before it's declared (which is valid in JavaScript but harder to read)
- Many files mix required library statements with require statements for local resources.  We typically prefer an order to these with libraries being loaded first.  We also don't want require statements embedded in the document; they should all be listed at the head of the document for maintainability.

app/lib/Actions.js
krishmaj: Noted - I'll go back through my comments and remove some, like this, which in hindsight are unnecessarily obvious.

app/lib/Actions.js
krishmaj: Changed.

app/lib/Actions.js
krishmaj: Changed - thanks for the explanation.

app/lib/Actions.js
krishmaj: Changed here, and hopefully I caught any other instances in this file as well.  Will change in other files as I go.

app/lib/Actions.js
krishmaj: Moved in this file, and will do same in other files as I go.

app/lib/Actions.js
krishmaj: Changed in this file, and will do same in other files as I go.

app/lib/Actions.js
krishmaj: Changed.

app/lib/Actions.js
krishmaj: Changed.

app/lib/Actions.js
krishmaj: Changed the bit inside the brackets, but it has to be a `let`, not a `const`.

app/lib/Actions.js
krishmaj: Changed.

app/lib/Actions.js
krishmaj: As above, will clean up my comments in a later PR.

app/lib/Actions.js
krishmaj: Changed.

app/lib/Actions.js
krishmaj: Changed, and removed the lines below it.

app/lib/Actions.js
krishmaj: Changed all short conditions that I could find in this file ... will also change elsewhere as I go.

app/lib/Actions.js
krishmaj: Changed.

app/lib/Actions.js
krishmaj: @micchickenburger I think you're probably right, but I'm not sure what the correct way is.  I do need to declare `typeformId` as a field in the `match_condition` before the related `if` statement immediately below (otherwise it gives an error saying that it cannot find `match_condition.typeFormId`, which makes sense of course).  So should this line just be `typeFormId = {}`?

app/pug/list_components.pug
krishmaj: I'll make this change as part of the components code cleanup which will be a future PR (is second in line on high priority issues).

app/pug/list_jobTypes.pug
krishmaj: Both this file and the one below are going to be completely removed in the next PR in about a week, so I'm leaving this as-is for now.

app/routes/actions.js
krishmaj: @micchickenburger Like `const default_form_schema = require(JSON.parse(require('fs').readFileSync('./schemas/default_form_schema.json')));` ?  And also, should the internal `require('fs')` in this line be its own separate line?

app/routes/actions.js
krishmaj: Changed.

app/routes/actions.js
krishmaj: Changed.

app/routes/actions.js
krishmaj: Changed this and all similar instances in this file to `500`.

app/routes/actions.js
krishmaj: Yes, you're correct - changed this and all similar instances in this file.

app/routes/start.js
krishmaj: Removed here and in the other `routes` files in this PR.  Will remove from others separately.

app/routes/start.js
krishmaj: Changed.

app/routes/start.js
krishmaj: @micchickenburger Not 100% sure.  Would it be better to do something like `if(!run.hasOwnProperty('user'))` for the previous `if` statement, so that we make sure that `req.user` is definitely defined at this line?  Similarly to one of your other comments, I think as it is right now, that previous `if` would still return `true` if `req.user` is undefined?

app/routes/tags.js
krishmaj: Changed.

app/routes/tags.js
krishmaj: Removed.

app/routes/tags.js
krishmaj: Changed here and in the other `routes` files in this PR.

app/routes/tags.js
krishmaj: Changed.

app/routes/tags.js
krishmaj: Changed.

app/static/pages/action_editTypeForm.js
krishmaj: Changed, and redone indentation of entire file accordingly.

app/static/pages/action_editTypeForm.js
krishmaj: Changed here and all other instances in file.

app/static/pages/action_editTypeForm.js
krishmaj: Added in all relevant places in the file.

app/static/pages/action_editTypeForm.js
krishmaj: Every time I looked at this file (and the others like it), I always had the feeling that something was wrong with it, but I could never quite put my finger on it ... bugged the hell out of me!  Yeah - a non-JSON file written as if it was JSON.  Changed.

app/static/pages/action_listTypes.js
krishmaj: @micchickenburger Added the missing semicolons.  Will look into changing from jQuery to native code separately.  Yes - this is the only input that is ever expected to be on this page.  But I think I see where you're going with it, and if in the future we add some additional input element, I will revisit this.

app/static/pages/action_specComponent.js
krishmaj: Changed.

app/static/pages/action_specComponent.js
krishmaj: Changed.

app/static/pages/action_listTypes.js
micchickenburger: I hope you remember to.  There might be a situation where you add an input element without revisiting this code, and the code might work during development but will fail in other browsers or environments where the DOM order differs.

app/lib/Actions.js
micchickenburger: I don’t think that’s right.  It’s a JavaScript object so members can be added any time in the future.

app/routes/actions.js
micchickenburger:
```suggestion
const { readFileSync } = require('fs');
const default_form_schema = JSON.parse(readFileSync('./schemas/default_form_schema.json'));
```

PR #38, opened 2022-06-16
Follow up changes for code quality in Actions and miscellaneous files:

krishmaj:
- adding commas after the last element of lists
- changing usage of var to let or const as appropriate
- removing usage of use strict in modules
- removing quotations around keys in non-JSON objects
- changed to backtick notation and consistent use of single quotes
- removed unnecessary usage of jQuery, replaced by native JavaScript equivalents
- moved module exports to ends of corresponding files, and re-ordered require statements to be library first, then local resources
- cleaned up short conditional statements to be only single line


micchickenburger (approved):
Thanks for cleaning all this up!

PR #43, opened 2022-07-07
Miscellaneous Changes as Requested

krishmaj:


krishmaj:
Forgot to mention that this also includes changing the batch sub-components links filename to use the batch order number, not the batch name.

Also, I've made no effort to do any code quality changes on the `/formio` file ... the whole lot of them need thoroughly cleaning up and fixing, which I'll do at a later date.

bjrebel (approved):
These changes look good to me.

PR #46, opened 2022-07-25
Increasing maximum batch size as requested

krishmaj:


bjrebel (approved):
Simple change to account for board batch sizes being larger than 50

PR #47, opened 2022-07-25
Increasing maximum batch size as requested

krishmaj:


bjrebel (approved):
Simple change, easy to approve

PR #49, opened 2022-07-27
Workflows - adding ability to reorder path steps when editing type form

krishmaj:


bjrebel (approved):
A useful change to reorder the steps in a workflow.

PR #50, opened 2022-07-27
Miscellaneous code quality and cleanup changes:

krishmaj:
- cleaned up common schema file (insertion and validity object generation) and removed associated now unneeded arguments from function calls
- fixed bugs in action and workflow type lists which was causing an error when no component types had been set
- changed double quotes to single quotes in API routes files
- added correct permission to API route for editing type forms
- removed Messages from the Developers box on homepage ... not really necessary and a pain to always keep updated

### DIFF
--- a/app/lib/commonSchema.js
+++ b/app/lib/commonSchema.js
@@ -1,64 +1,47 @@
-"use strict";
 
-const ObjectID = require('mongodb').ObjectID;
-var MUUID = require('uuid-mongodb');
-const logger = require('./logger');
+/// Insertion data for a single record
+function insertion(req) {
+  // Check that the minimum required information has been provided for a record's insertion data
+  // This is simply user profile information (stored in 'req.user')
+	if(!req) throw new Error(`commonSchema::insertion() - the 'req' object has not been specified!`);
+  if(!req.user) throw new Error(`commonSchema::insertion() - the 'req.user' has not been specified!`);
+
+  // Retrieve the specific user information fields
+	const {user_id, displayName, emails} = req.user;
+  
+  // Return the required information using the appropriate 'insertion.XXX' field names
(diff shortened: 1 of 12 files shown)


app/lib/commonSchema.js
```diff
@@ -1,64 +1,47 @@
...
-{
-	if(!req) throw new Error("No request object in commenSchema/insertion");
-	if(!req.user) throw new Error("No user info");
-	const {user_id, displayName, emails} = req.user;  // equivalent to user_id = user.user_id, etc
-    
+/// Insertion data for a single record
+function insertion(req) {
+  // Check that the minimum required information has been provided for a record's insertion data
+  // This is simply user profile information (stored in 'req.user')
+	if(!req) throw new Error(`commonSchema::insertion() - the 'req' object has not been specified!`);
```
bjrebel, line 6: The formatting looks strange here - why the extra tab?

app/lib/commonSchema.js
```diff
@@ -1,64 +1,47 @@
...
+  if(!req.user) throw new Error(`commonSchema::insertion() - the 'req.user' has not been specified!`);
+
+  // Retrieve the specific user information fields
+	const {user_id, displayName, emails} = req.user;
```
bjrebel, line 10: why the extra tab?

app/lib/commonSchema.js
```diff
@@ -1,64 +1,47 @@
...
+  
+  // Return the required information using the appropriate 'insertion.XXX' field names
 	return {
```
bjrebel, line 13: looks like extra tabbing here too.

bjrebel (approved):
Some tabbing to maybe remove, but code looks fine otherwise.

PR #54, opened 2022-07-29
Adding ability to perform a new action of the same type ...

krishmaj:
...  (component to be specified) directly from the record page of any already performed action


bjrebel (approved):
Changes look fine to me.

PR #55, opened 2022-08-02
Removing use of caches ...

krishmaj:
... for retrieving lists of component count by type and type form tags ... might solve the issue of duplicated UKIDs on geometry boards.


bjrebel (approved):
Changes look good

PR #59, opened 2022-08-08
M2M and Display Tweaks

krishmaj:
Tweaking layouts for 1920 x 1080 resolution ... previous layout was built on 2k screen, so some things were wider than they should be for the more common resolution.  Also applying autopep8 formatting to the M2M Python scripts, and adding exception handling for socket timeouts.


bjrebel (approved):
Changes look fine to me.

PR #60, opened 2022-08-12
Searching - entity record by UUID or ID:

krishmaj:
- cleaned up code for Formio component UUID input box, and combined old QR_Component code in, since there is no reason for it to be a separate thing
- added new Formio action and workflow ID input boxes and removed old TestID box (code and references in Formio register)
- cleaned up and tested autocomplete library functions for all entity types
- added action and workflow ID search boxes to previous Search Component by UUID page, and renamed to generalise


bjrebel (approved):
These changes look sensible to me.

PR #62, opened 2022-08-16
Relaxing short UUID regex ...

krishmaj:
... from exactly 22 characters to between 20 and 22


bjrebel (approved):
Very sensible changes.

PR #63, opened 2022-08-16
Diagnostic code for retrieving and displaying the maximum typeRecordN…

krishmaj:
…umber value per component type ... temporarily in place for fixing the UKID discrepancy bug


bjrebel (approved):
These look fine, let's hope they solve the problem

PR #66, opened 2022-08-30
Code Cleanup - pugs, statics, css and routes:

krishmaj:
- cleaned up and made consistent all pug files
- cleaned up and made consistent all css files, and removed unused sections
- removed the 'default_form_schema' JSON file - there's no need to have it as a separate file, since the default schema is always empty.  Routes that use it now have an empty schema hardcoded.
- cleaned up scss file, and removed unused sections
- removed unused image file, and confirmed that all remaining images are being used somewhere
- moved 'qrcodegen' library file into 'static/js' directory with the other externally-sourced JS code
- renamed 'generate_qr_code' static JS file to be consistent with other files in 'static/js' directory, and cleaned up file
- cleaned up and made consistent all static page files
- added more explicit displayed messages to Search Description and Search by Visual Inspection / Order Number pages regarding boards not being displayed in the search results if they haven't had a Visual Inspection performed
- cleaned up comments in all routes (API and non-API)
- removed old and unused test scripts (these were related to the old M2M setup, no longer relevant)


bjrebel (approved):
Changes look fine to me.

PR #70, opened 2022-09-08
Library Files Cleanup ...

krishmaj:
- cleaned up comments on all 'app/lib' files
- removed 'email' and 'database' lib files ... former wasn't being used anywhere at all, and latter was previously being used as part of the old M2M test scripts
- removed 'Cache' lib file, and changed type forms list function (in 'Forms') to get lists directly from the database, rather than via caching
- removed diagnostic function for maximum component type record numbers, along with usage in route and pug file (no longer needed since the UKID issue seems to have been solved)
- removed remaining references to caching from main app file and one route comment


bjrebel (approved):
All reasonable changes, appreciate keeping comments up to date.

PR #71, opened 2022-09-08
Adding explicit static route to short-uuid resource file

krishmaj:


bjrebel (approved):
Looks good, let's hope that is all we need.

PR #72, opened 2022-09-08
Another attempt to fix the short-uuid resource file problem

krishmaj:


bjrebel (approved):
Definitely worth a shot

PR #81, opened 2022-09-13
Simpler Server-side Short UUIDs and Actions Bug Fix

krishmaj:


bjrebel (approved):
Makes sense to only allow full UUIDs for now as there are several spots in the code that could otherwise fail.

micchickenburger:
I think this approach violates DRY:  Don’t Repeat Yourself.

If function A takes parameter of type UUID and returns an object of type Component, and function B does the same then you have repeated yourself.  It is a property of UUID that it have variable length.

PR #89, opened 2022-09-14
Edge case fix for shipment with no boards ...

krishmaj:
- where a board shipment has a board entry but no board UUID, UUID and UKID are now set to 'none' and displayed as such
- why would you make a shipment that doesn't have any boards ... why not just wait until you have at least one to put in


bjrebel (approved):
looks good

PR #91, opened 2022-09-15
Cleanup of static JS, Formio and top-level files

krishmaj:


bjrebel (approved):
clean up looks good

PR #92, opened 2022-09-21
Packages Cleanup (part 1):

krishmaj:
- changed all library and route files to use relative internal paths (from absolute), and removed corresponding use of the 'app-modules-path' package
- removed use of 'moment' package from page JS file, since a standard JS date object functions exactly the same
- changed all package libraries and stylesheets that were previously imported via a static 'dist' filepath to use external CDN sources instead
- removed all usage of static 'dist' filepaths from the DB code


bjrebel (approved):
Changes look good to me

PR #93, opened 2022-09-21
Miscellaneous requested changes:

krishmaj:
- increased number of displayed components, actions and workflows on list pages from 100 to 200
- moved 'Perform a New Action of this Type' button from bottom of action information page to be inline with action type at top
- added a 'Create a New Component of this Type' button to component information page, inline with component type at top
- added display section for the 'data.typeRecordNumber' field (as 'Factory ID') for 'Grounding Mesh Panel' type components (no display for other types)

### DIFF
--- a/app/pug/action.pug
+++ b/app/pug/action.pug
@@ -22,7 +22,14 @@ block content
     .vert-space-x1
       dl
         dt Action Type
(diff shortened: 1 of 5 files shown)


bjrebel (approved):
Changes look fine.

PR #95, opened 2022-09-24
Cleaning Up Package Dependencies

krishmaj:
Please see individual commits for details.  The remaining packages that still need looking into are `express-compile-sass` (no development since 2018) and `node-sass` (deprecated, alternatives are available), but since these are very specifically for sass compilation, we can deal with them at a later date.


bjrebel (approved):
Nice job cleaning things up

PR #96, opened 2022-09-28
M2M Client-side Script changes and fixes ...

krishmaj:
- optimised access token usage ... now a single token can be used for multiple calls to the various DB API routes, rather than one token per call
- changed short UUID conversion to use the more complete API route (which includes checking the full UUID against existing DB records)
- removed the now unused less complete short UUID conversion API route
- updated M2M README and in-code comments
- added new template explicitly for short UUID conversion, and updated existing templates
- fixed bug for non-existent components in board shipment information page
- removed 'Paste JSON Schema from Clipboard' button from editing type form pages ... this no longer works in Firefox, and it will be confusing to keep it available only for other browsers


bjrebel (approved):
these changes look ready for staging tests

PR #97, opened 2022-09-30
Requested changes to M2M scripts and Perform Action page ...

krishmaj:
- component name is now displayed at the top of the 'Perform Action (Specified Component)' page, in addition to the existing action type and component UUID
- added client-side M2M functions for retrieving component and action records
- added new template for retrieving records
- component creation, action performance and record editing client-side M2M functions now all return the entity UUID or ID
- updated M2M README


bjrebel (approved):
These look fine to me.

PR #104, opened 2022-10-20
Additional check on shipment vs. reception destination for new search query

krishmaj:


bjrebel (approved):
A straightforward change

PR #106, opened 2022-11-07
Requested changes to Geometry Board Shipment search ...

krishmaj:
- added option to query against presence or lack of shipment comment
- split search more explicitly into received vs. unreceived shipments ... makes the library function a little easier to handle
- added corresponding new query box to interface page, and associated logic and argument to page-level JS


bjrebel (commented):
Looks reasonable to me

bjrebel (approved):

PR #109, opened 2022-11-28
Miscellaneous Interface Tweaks

krishmaj:
Tweaks to various interface pages, based on feedback and suggestions from users.  Please see individual commits for details.


bjrebel (approved):
These look good to me

PR #110, opened 2022-11-28
Bug fixes for earlier changes ...

krishmaj:
- if no component name is available, the UUID will now be shown in the component name column on the list of components page
- fix to account for not finding a component related to a workflow ... in which case, no attempt will be made to find a (non-existent) component name


bjrebel (approved):
Looks good

PR #112, opened 2022-12-07
Changes to Action Type Forms list page ...

krishmaj:
- type forms are now grouped according to their recommended component type
- removed now unneeded commended component type column
- groups are displayed in collapsible panels, which are collapsed by default
- added new library function for retrieving grouped list of type forms (do not want to modify the existing, non-grouped function because that is used in other places)


bjrebel (approved):
These changes are fine to me

Issue #25, opened 2022-04-29
OAuth Error on User Login


micchickenburger:
```
InternalOAuthError: Failed to obtain access token
--
  | at Strategy.OAuth2Strategy._createOAuthError (/app/node_modules/passport-oauth2/lib/strategy.js:408:17)
  | at /app/node_modules/passport-oauth2/lib/strategy.js:175:45
  | at /app/node_modules/oauth/lib/oauth2.js:191:18
  | at ClientRequest.<anonymous> (/app/node_modules/oauth/lib/oauth2.js:162:5)
  | at ClientRequest.emit (node:events:527:28)
  | at TLSSocket.socketErrorListener (node:_http_client:454:9)
  | at TLSSocket.emit (node:events:527:28)
  | at emitErrorNT (node:internal/streams/destroy:157:8)
  | at emitErrorCloseNT (node:internal/streams/destroy:122:3)
  | at processTicksAndRejections (node:internal/process/task_queues:83:21)

```

micchickenburger:
This occurs on first login for a user

krishmaj:
With @micchickenburger 's addition of debug messaging (#113), I now get the following:

```
ERROR: {"name":"InternalOAuthError","message":"Failed to obtain access token","oauthError":{"errno":-3001,"code":"EAI_AGAIN","syscall":"getaddrinfo","hostname":"dunedb-prod.us.auth0.com"}} undefined undefined
```

when doing a new login to production.  Would seem to emanate from https://github.com/DUNE/dunedb/blob/eea01570d6d9c72aabd48b7f17a2f9e297ef5480/app/lib/auth.js#L26

krishmaj:
@micchickenburger Your suggested workaround seems to have at least stopped the error from impacting users, at least from my (admittedly low usage) testing.  We've asked the users to keep an eye out for any re-occurrences.  And we should keep this issue open as a reminder that we still need to, at some point, address it properly with a refactoring of the entire authentication system.

PR #118, opened 2023-01-22
Temporary workaround for Auth0 callback error ...

krishmaj:
- see issue #25
- workaround implements a manual redirection back to the homepage when a callback error is encountered


bjrebel (approved):
looks fine to me.

PR #119, opened 2023-01-31
Extending geometry board visual inspection search ...

krishmaj:
... to allow matching against issues

- new dropdown menu on interface page for optionally selecting an issue to refine the search with
- added corresponding page level code and changes to displayed output
- added new optional parameter to search route
- reworked search library function to allow for an optionally specified issue string
- updated search description page


bjrebel (approved):
Looks good

PR #121, opened 2023-02-06
Extending geometry board search by order number ...

krishmaj:
- search now includes order numbers associated with batches of returned boards
- extended library function to search returned board batches if no match is found in intake board batches
- for each geometry board in a matching batch, search now returns action data about the single latest visual inspection (since boards may now have multiple inspections performed ... one on intake and one on return)
- reverted displayed 'all' back to 'any' in issue selection on inspection search page ... the former makes more sense in context


bjrebel (approved):
Looks fine to me

PR #125, opened 2023-02-15
New M2M functionality - listing components and actions of a single type …

krishmaj:
- new API routes to return lists of component UUIDs and action IDs for a specific component or action type form ID respectively
- new backend M2M functions to access new routes
- new template to illustrate new functionality, and updated M2M README


bjrebel (approved):
These look good to test

PR #131, opened 2023-03-09
Changes to actions in component summary documents ...

krishmaj:
- actions are now divided into non-conformance, workflow and other ... each section indicated by the blue banner on the left of the action
- actions within each section are now displayed chronologically, i.e. earliest first (was previously reverse chronologically)


bjrebel (approved):
Changes look fine to me.

PR #132, opened 2023-03-20
Miscellaneous changes from 2023-03-20 APA DB meeting ...

krishmaj:
- changed component summary action banner to show 'other action' for better clarity (was previously just 'action')
- changed order of search pages in sidebar to group UUID related searches together, as requested
- changed order of search page descriptions to match new sidebar order


bjrebel (approved):
Changes look good.

PR #134, opened 2023-03-21
Small display fixes:

krishmaj:
- sub-component table column header on batch component information pages now shows either 'UK ID' or 'UW ID' depending on the type of sub-component
- small fix for misplaced horizontal line on search description page (left over from previous commit)


bjrebel (approved):
looks good to me.

PR #136, opened 2023-03-28
Small changes as requested:

krishmaj:
- action section titles in component summary now show the action type name
- components with DUNE PIDs as names now have those automatically set upon component creation, in both batch and non-batch


bjrebel (approved):
Discussed via zoom

PR #140, opened 2023-04-17
Component Name Display on Workflow Information Page

krishmaj:
If it exists, the component name is now displayed in the workflow information page's path table ... if no name exists, the component UUID is displayed as previously.


bjrebel (approved):
good change to have

PR #141, opened 2023-04-21
Miscellaneous fixes and tweaks ...

krishmaj:
- attempting to fix the displayed order of action type groups and types within groups on the action types list page
- bug fix for disappearing data when editing an action whose type form uses a NumberArray Formio component
- small tweaks to the graph and histogram displayed as part of the NumberArray Formio component
- re-enabled ability to edit actions whose type form uses a NumberArray Formio component (previously disabled due to aforementioned bug)


bjrebel (approved):
Looks reasonable to me

PR #145, opened 2023-05-09
Attempting fix for batch sub-components going missing ...

krishmaj:
- changed component creation/editing page so that all sub-component data is sent to the server in one go, and the server now handles submitting each one individually to the DB
- this should reduce the frequency of API calls to the DB pod
- added new corresponding route specifically for submitting sub-components, and added route to JEST script (still doesn't actually work, but may as well keep it updated)


bjrebel (approved):
Looks like a good change to make

PR #146, opened 2023-05-09
Merging Batch Components Fix from Staging

krishmaj:


bjrebel (approved):
Changes already tested in staging

PR #149, opened 2023-06-19
Changes to M2M application for wire tensions uploading ...

krishmaj:
- new backend function for extracting tension measurements from a specified .csv file and converting to a Python list for DB upload
- this should be followed by a standard M2M call to either perform a new action or edit an existing one as appropriate
- updated README with new section describing use of the M2M application with wire tensions


bjrebel (approved):
I think these are probably fine.  We want to double check that the segment numbering is consistent between the DWA and this code.

PR #151, opened 2023-07-18
New search for actions by referenced component UUID

krishmaj:
- new library function, interface route, interface page and API route added new search to sidebar
- updated search descriptions and jest routes file


bjrebel (approved):
looks good

PR #153, opened 2023-07-26
Attempting Bug Fixes

krishmaj:
- Potential fix for UUIDs disappearing with sequential entries using QR code scanning ... these changes make QR code and manual entry fully consistent, but I don't know if this actually fixes the problem
- Reworking Search for Geometry Board Shipments by Reception Details - search will now only be performed when the user explicitly presses the 'Perform Search' button ... this should prevent multiple unwanted searches being performed simultaneously, and then displaying unwanted results


bjrebel (approved):
Looks like this could solve the issue

PR #154, opened 2023-08-07
Fixes to the way the tension data CSVs are read

krishmaj:
These are mostly needed due to the CSVs being slightly different depending on the layer


bjrebel (approved):
Looks fine to me

PR #155, opened 2023-08-09
Miscellaneous Bug Fixes

krishmaj:
All previously tested on staging:

- fixed previously entered UUIDs being deleted when scanning multiple QR codes
- fixed existing "Search for Board Shipments by Reception Details" results being overwritten after delay
- fixed incorrect expected CSV formatting when reading tension measurements data


bjrebel (approved):
All changes previously tested on staging.

PR #157, opened 2023-10-13
Small addition to M2M frame survey results upload script

krishmaj:
New field in type form, plus corresponding small formatting fixes


bjrebel (approved):
looks good to me

PR #165, opened 2023-11-22
Small updates ...

krishmaj:
- fixed remaining instances of old search page URLs ... doesn't seem to be broken even using the old URLs, but should keep things updated anyway
- added winder head information to Assembled APA executive summary


bjrebel (approved):
looks good to me.

PR #169, opened 2023-11-28
Adding ability to track board installation ...

krishmaj:
- geometry board location is now set to 'installed on APA' for all boards in a board installation action, upon submission of such actions
- new backend actions route and associated page level JS function
- added new 'location' to available locations when searching for boards by location


bjrebel (approved):
Looks good to me.

PR #171, opened 2023-12-20
Miscellaneous changes relating to APA 8 Executive Summary ...

krishmaj:
- added dictionary to convert winding head field value from unformatted to formatted
- made it more clear when there are no wire-related NCRs (as opposed to if that section of the summary has just failed to generate)
- same as above for the other NCR sections as well
- fixed the incorrectly shown tension measurement threshold lines
- fixed a bug that prevented a tension measurements action being edited if no previous tensions were recorded  ... library function now calculates re-tensionings if previous version of the record contains tension measurements
- renamed certain variables to make it clear that re-tensionings are done by comparing current and immediately previous versions of the action
- action information page now shows re-tensioning comparison tables if such data exists in the record (related to bugfix in library function)
- updated M2M tension measurements upload script and backend function for new layer template spreadsheets ... testing using basic G layer example

### DIFF
--- a/app/lib/Components_ExecSummary.js
+++ b/app/lib/Components_ExecSummary.js
@@ -278,6 +278,7 @@ async function collateInfo(componentUUID) {
 
   const dictionary_locations = {
     daresbury: 'Daresbury',
+    chicago: 'Chicago',
   };
 
   const dictionary_systems = {
--- a/app/lib/Components_ExecSummary.js
+++ b/app/lib/Components_ExecSummary.js
@@ -264,6 +264,18 @@ async function collateInfo(componentUUID) {
     usWinder1: 'US Winder 1',
   };
 
+  const dictionary_heads = {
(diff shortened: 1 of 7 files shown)


app/lib/Components_ExecSummary.js
```diff
@@ -264,6 +264,18 @@ async function collateInfo(componentUUID) {
...
+    uk4: 'UK 4',
+    uk5: 'UK 5',
+    uk6: 'UK 6',
+    uk7: 'UK 7',
+    us1: 'US 1',
+    us1: 'US 2',
+  };
+
   const dictionary_locations = {
     daresbury: 'Daresbury',
```
bjrebel, line 280: Shouldn't we add in Chicago here too?

krishmaj: Resolved

bjrebel (approved):
Changes look good to me

PR #173, opened 2024-01-03
Updating frame vertical inspection related code ...

krishmaj:
- split M2M analysis results extraction function into two - separately for envelope and planarity, since they will likely come from different data files
- reworked new separate planarity analysis results extraction function for use with standardised .csv results files
- corresponding updates to template upload script, plus renaming both M2M files to better clarify that they deal with VERTICAL inspections (not horizontal)
- reworked action information page pug template ... removed table of raw datapoints, checking that the results exist before attempting to display them, better formatting
- renamed instances of the vertical frame inspection action type form name for better clarity of purpose


bjrebel (approved):
These changes look good to me.

PR #174, opened 2024-01-05
Semi-freeform Workflows and Executive Summaries

krishmaj:
First Commit:
- workflows may now contain a section in which the steps may be done in any order ... selection of freeform steps is hardcoded and dependent on the workflow type
- step index is now propagated through the interface, routes and library, and library function now explicitly sets the result of the specified step (previously would assume to set the result of the first step with no result)
- reworked workflow information page accordingly ... new conditionals according to step result, position of step in workflow, whether step is freeform or not
- cleaned up some unnecessary code relating to submitting components that will never be part of workflows (previously would still check if they were)

Second Commit:
- overhaul of how workflow completeness is determined ... individual actions now have their own completeness tracked (via checkbox), and the workflow will now check the individual actions directly for their completeness
- individual action completeness is now displayed on the workflow information page, and workflow completeness derives from if all individual actions are explicitly complete or not
- workflow completeness no longer stored in the workflow record, but instead determined only as and when required (only on two pages of the interface)
- assembled APA executive summaries now display more user-friendly information if and when actions have not yet been performed or completed ... default field value when QC review personnel are missing, and links disabled and relabelled when action IDs are missing

### DIFF
--- a/app/lib/Components_ExecSummary.js
+++ b/app/lib/Components_ExecSummary.js
@@ -1,5 +1,6 @@
 const MUUID = require('uuid-mongodb');
 
+const Actions = require('./Actions');
 const Components = require('./Components');
 const { db } = require('./db');
 const Search_ActionsWorkflows = require('./Search_ActionsWorkflows');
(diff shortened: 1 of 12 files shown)


bjrebel (approved):
Changes look good to me.

PR #175, opened 2024-01-25
Fixing probable bug in Executive Summary code

krishmaj:
Length of mesh panel NCRs array was not being checked before potentially being used.
Also improved the if-statement logic of the "Other NCRs" section of the summary ... previously would only check if ANY of the sub-sections had information to show, now checks them individually as well.


bjrebel (approved):
Good catch

PR #176, opened 2024-01-26
Still tracking bug in Exec. Summaries ...

krishmaj:
- temporarily disabling code for determining workflow status on summary page - this may be the source of the bug


bjrebel (approved):
those lines are def commented out

PR #177, opened 2024-01-28
Fixing workflow status code in Exec. Summaries ...

krishmaj:
- found source of previously noted bug, fixed
- also found a couple of other bugs in inconsistent field names, fixed


bjrebel (approved):
Let's see how much further we get now.

PR #180, opened 2024-02-03
Attempting to reduce memory usage when sorting and returning list of all actions

krishmaj:
- removed unnecessary group fields (both components and actions ... left over from old work)
- reordered limit and skip optional queries relative to second sorting


bjrebel (approved):
changes look reasonable

PR #182, opened 2024-02-05
Bug Fixes for Memory Limit and non-existent Inspection Results ...

krishmaj:
- trying a different approach to fix the memory limit exceeded bug when sorting too many actions - new flag allows MongoDB to use local temporary files to help
- bug fix for displaying non-existent vertical inspection results ... previously code assumes that all results are available, but some are missing for older frames.  New code will now show empty lines where appropriate, instead of crashing


bjrebel (approved):
approving to try it out

PR #186, opened 2024-02-10
Another attempt to fix Memory Limit bug ...

krishmaj:
- trying use of MongodB 'project' aggregation stage to reduce the amount of information being passed to the 'sort' stage
- removed 'allowDiskUse' flags recently added ... don't appear to be doing anything, and it is actually enabled by default in MongoDB v6.0, so wouldn't be needed anyway in the future


bjrebel (approved):
Hopefully this attempt works

PR #188, opened 2024-02-14
Small changes to frame envelope analysis related code ...

krishmaj:
- action information can now display table of envelope analysis results if some parameters have no given tolerance (previously would crash)
- better positioning of terminal output messages for successful envelope and planarity analysis result extraction


bjrebel (approved):
changes look fine to me.

PR #190, opened 2024-02-15
Reverting some previous changes ...

krishmaj:
- putting sort operation by last edit date back to being done before limit operation
- was previously like this, but I changed the operation order in one of the earlier attempts to fix the exceeded memory bug ... forgot to revert
- removing unused skip operations from component, action and workflow list library functions - this has never been used, and never will


bjrebel (approved):
Apparently this is the preferred way of doing things

PR #192, opened 2024-02-15
Safeguarding against future 'Exceeded Memory Limit' bug reoccurrence ...

krishmaj:
- applied 'project' MongoDB operators to component and workflow list library functions, to reduce the amount of information being passed to later 'sort' operators
- equivalent to recently added operator on action list function, which solved the bug there


bjrebel (approved):
seems like a good plan to future-proof things

PR #194, opened 2024-02-17
Updates to and cleaning up of M2M scripts ...

krishmaj:
- consolidated extraction functions for vertical and horizontal frame inspection results into a single file ... no real reason to have them separate
- renamed dictionary of M4 hole measurements to be explicit as to contents ... just in case some other horizontal inspection results might be included in the future
- moved and re-worded some terminal output messages to be better placed and clearer about where they emanate from
- renamed target field in action record for storing M4 hole measurements to be explicit as to contents ... just in case some other horizontal inspection results might be included in the future
- removed 'Perform New Action' parts of tension measurements and frame inspection results upload scripts ... these action types are part of workflows, and will therefore always be initially performed through the web interface, never through an M2M script.  Hopefully removing this from the scripts will deter users from accidentally making standalone versions of these actions.
- updated README with general section on uploading data from external files


bjrebel (approved):
All reasonable changes

PR #196, opened 2024-02-20
Update to numbers of bad solders and replaced wires in Exec. Summary ...

krishmaj:
- counters for the number of bad solders and replaced wires per wire layer now based on how many comma-delimited values are entered
- code accounts for if the records contain numbers (old style) or strings (new style) entries ... saves having to change every single existing record
- numbers are now passed directly to the interface, rather than being determined as the length of lists in the latter


bjrebel (approved):
looks good to me.

PR #197, opened 2024-02-20
Update to numbers of bad solders and replaced wires in Exec. Summary

krishmaj:


bjrebel (approved):
reviewed on staging already

PR #198, opened 2024-02-27
Pre-emptive bug fix for replaced wires and bad solders ...

krishmaj:
- currently, the code checks if there are three empty strings - 'side', 'layer' and 'boardLocation' - in the first entry of the 'replacedWires' or 'badSolderJoints' field
- if so, the entire field is replaced by an empty list, since the empty strings would indicate that no wires or solders have actually been entered
- HOWEVER, because the 'layer' sub-field has a default value in the type forms, this will never be an empty string ... and so there will always be an entry in the list, even if no wires or solders have indeed been entered
- the check on the 'layer' string has now been removed ... the 'side' and 'boardLocation' string checks remain, since they can be empty (and will be, if no wires or solders have been entered)


bjrebel (approved):
Looks good to me

PR #199, opened 2024-02-27
Pre-emptive bug fix for replaced wires and bad solders

krishmaj:


bjrebel (approved):
Approved previously on staging

PR #200, opened 2024-03-08
Changing development instance deployment file to use MongoDB v6.0

krishmaj:


bjrebel (approved):
Changes clearly needed for upgraded DB

PR #202, opened 2024-04-05
Requests relating to Populated Board Kits

krishmaj:
- populated board kit constituent components (CR boards, G-Bias boards, SHV boards and cable harnesses) can now be searched for by type and number
- they can also now be received at a specified location (using a 'Populated Board Kit Reception' action), and then searched for by reception location


bjrebel (approved):
Looks good to me.

PR #203, opened 2024-04-07
Miscellaneous requested changes ...

krishmaj:
- changed to using a local copy of 'jsQR.min.js' ... this will hopefully fix the recent issue with Chrome / Edge not being able to access a connected webcam for QR code scanning
- removed the ability to add an 'edited action' comment when uploading tension measurements via the M2M script ... I'm told that this is never actually used (comments are always added through the interface), so this just cleans things up a little
- changed the logic that is used to determine a list of most recently changed tensions ... the DB will now look back through ALL versions of the action until it finds one with different tensions to the current (previously would only look at the immediately preceding version)
- new GUI for uploading tension measurements to the DB ... this is an alternative to using the original Python script on the command line
- changed tension measurements Python script to allow command line parameters ... if values are given, they will be used in place of any given in the script (if no parameters are given, usage remains as previously)
- new .bat file that allows the GUI to be used like a normal program (i.e. without needing to run the GUI itself from the command line)
- fixed small warning relating to string comparisons ... apparently 'is not' is valid in Python 3.11 (the version I use for development), but not in 3.12 (the version being used on the winders at Daresbury)


bjrebel (approved):
All good ideas

PR #208, opened 2024-05-02
Implementing location tracking for all components ...

krishmaj:
- all components are now assigned a default location at creation (specific location depends on component type)
- location information for certain component types can be changed through submitting certain actions or creating certain other component types
- moved component name and location assignment to centralised block in server side library function (was previously done in various places on the client side)
- made server/client side codes for actions and workflows consistent with newly re-organised component codes (only small changes needed)
- new centralised dictionary of location key/value pairs, which is passed through to the interface pages as required (for consistent displayed locations across the entire DB)
- added new pug mixin for displaying only the date (not the time as well), and renamed the existing mixin to explicitly show that it displays the date AND time
- a component's current location is now displayed on its information page, along with the date at which it moved there
- if a component is installed on an APA (geometry boards and grounding mesh panels), the specific APA will be recorded in the component's location information
- update location search interface pages with additional potential locations that have been recently added to various action type forms

### DIFF
--- a/app/lib/Actions.js
+++ b/app/lib/Actions.js
@@ -36,7 +36,7 @@ async function save(input, req) {
   newRecord.recordType = 'action';
   newRecord.actionId = new ObjectID(input.actionId);
   newRecord.typeFormId = input.typeFormId;
-  newRecord.typeFormName = input.typeFormName || typeForm.formName;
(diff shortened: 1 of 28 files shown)


bjrebel (approved):
changes look reasonabler

PR #209, opened 2024-05-05
Implementing location tracking for all components

krishmaj:
Please see commit for details


bjrebel (approved):
Previously checked with staging.

PR #210, opened 2024-05-16
Updates to Searches by Location ...

krishmaj:
- all searches now use the centralised list of possible locations for consistent displayed names
- changed the previous 'Search for Assembled APA by Location' to make it more explicit that this is related to the APA's PRODUCTION location (and number), not necessarily its current location (though they may of course be the same)
- reduced mentions of 'reception location' in variable names and comments - since the component location is now generalised, and may not necessarily represent where a component was last received anymore
- added information about which APAs grounding mesh panels are installed on to 'Search for Grounding Mesh Panels by Location' search results
- reworked how information about which APAs geometry boards are installed on is generated ... no longer need to retrieve Board Installation actions, since the APA UUID is now stored in the board record itself

Note that the library function changes will need to be tested on Staging once that instance's records have been updated with new component information.

### DIFF
--- a/app/lib/Search_GeoBoards.js
+++ b/app/lib/Search_GeoBoards.js
@@ -2,14 +2,13 @@ const MUUID = require('uuid-mongodb');
 
 const Components = require('./Components');
 const { db } = require('./db');
-const Search_ActionsWorkflows = require('./Search_ActionsWorkflows');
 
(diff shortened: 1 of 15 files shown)


bjrebel (approved):
The changes look good to me.

PR #211, opened 2024-05-16
Additional changes to location searches ...

krishmaj:
- changed 'Search for Board Shipment by Reception Details' to correctly use the central list of locations ... was already being passed to the interface page, just not being used
- renamed some variables to avoid conflict with the global, immutable 'location' variable


bjrebel (approved):
Looks good to me

PR #212, opened 2024-05-17
Interface Improvements and Search Bug Fixes

krishmaj:
Interface Improvements
- [Sotiris] APA Frames and Assembled APAs are now shown in alphanumeric component name order on their single component type listing pages (listing pages for other component types and the general listing page remain in order of most recently edited component) 
- [Brian] all workflow listing pages now additionally show which action is to be completed next ... this may not work entirely well for APA Assembly workflows, since they have a free-form section at the start, but I can look into improving it after seeing it in action first 
- [Brian] component information pages for Board Shipments and Grounding Mesh Shipments now show each constituent component's part number 
- [Carlos] tables of changed tensions on Tension Measurements action information pages now display the wire or wire segment number, instead of the index (the data is still stored by index) ... note that the scatter plot of tension vs. wire is still axised by wire index (Formio limitation)
- [Sotiris] when displaying search results involving components, the component name (DUNE PID, UKID, etc.) will now always be displayed if available, and the UUID will only be shown if this is not possible

Search Bug Fixes
- some geometry boards and grounding mesh panels appear to be missing the 'reception.detail' field entirely ... this is causing MongoDB to set a certain array to 'null', and stopping the search library function from completing
- workaround is to first check if the 'reception.detail' field exists before attempting to pull its information ... not as efficient, since this has to be done outside the MongoDB aggregation, but it works
- fixed misnamed variable in Mesh Panel Search by Location / Part Number inteface page's JS code

### DIFF
--- a/app/lib/Components.js
+++ b/app/lib/Components.js
@@ -289,8 +289,12 @@ async function list(match_condition, options) {
     },
   });
 
-  // Re-sort the records by last edit date ... most recent first
-  aggregation_stages.push({ $sort: { lastEditDate: -1 } });
+  // Re-sort the records ... by (alphabetical) component name for APA frames and assembled APAs, or by last edit date (most recent first) for other component types
+  if ((match_condition) && (match_condition.formId) && ((match_condition.formId === 'APAFrame') || (match_condition.formId === 'AssembledAPA'))) {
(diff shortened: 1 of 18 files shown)


bjrebel (approved):
Looks good to me

PR #216, opened 2024-05-25
Bug fix for Assembled APA workflow not updating ...

krishmaj:
- workflow step 0 result (Assembled APA UUID) is now correctly updated after setting the Assembled APA and APA Frame locations upon Assembled APA component creation


bjrebel (approved):
Glad to have this fixed

PR #220, opened 2024-05-31
Updating APA Executive Summaries to use centralised personnel lists

krishmaj:
Plus a couple of additional small interface tweaks ... please see individual commits for details.


bjrebel (approved):
looks good

PR #222, opened 2024-06-06
Miscellaneous Changes ... Requests and Tweaks

krishmaj:
Additional small tweaks to APA Executive Summaries
- summary now links to APA frame INSTALLATION surveys action (not yet implemented outside dev instance, but to be soon)
- changed display of wire NCR information to match NCR action ... naming of FEMB-ASIC-ASIC field and order of fields in table

Changes to Formio graph components
- removed statistics output ... apparently nobody cares and it's just a waste of space
- changed scatter plot of wire vs. tension so that the x-axis now represents the wire or segment NUMBER (not index), starting at the correct value for each layer
- changed scatter plot of wire vs. tension so that the x-axis label is now correct for each layer (wire vs. segment)

Changes to Single-Type Workflow listing pages
- single type workflow listing pages now show workflows in order of reverse alphanumerical component name
- overall (all types) workflow listing page is unchanged - workflows are still shown in order of last edited, most recently first
- **this change has not been tested very well on Development instance (only a small number of workflows present) ... needs fully confirming on Staging**

Changes to lists of personnel
- added some more historical UK and US technicians
- changed lists of 'Tension Control' and 'Winder Maintenance' signoff personnel based on discussion with Radosav and Ed

 Overhauling Survey Upload M2M scripts
- changes originating from new methodology for frame survey analysis - separate intake and installation planarity analyses, plus intake M4 hole survey and installation envelope analysis
- changed to a single upload script for all surveys, pointing at two actions (intake and installation surveys)
- changed backend extraction functions to work with new spreadsheet for envelope and planarity analyses (M4 hole surveys spreadsheet is unchanged)
- changed action information page to work with new survey actions and formatting of survey results
- three cross corner measurements will be taken at DSM, not just one - added new backend function and modified M2M upload script accordingly
- modified action information page to show all three measurements ... displayed names match those in the frame documentation


bjrebel (approved):
These changes look fine. There may be a request to put the out of bounds counts back in by the compliance office or technical coordination, but let's cross that bridge when we come to it.

PR #224, opened 2024-06-26
Improvements to Info Page Layouts and Adding Replaced Wires Count Cross-check

krishmaj:
- adding page titles to those that don't currently have them, so that they are more easily identifiable in browser histories
- tweaking a couple of existing page titles for consistency

- adding framework for new Frame Installation Surveys Signoff personnel dictionary (to be populated at a later date)

Revamping component, action and workflow information pages
- more specific page titles, both in interface and for browser display / history - record type and type form name are now shown
- moved all interaction buttons (edit, JSON, etc.) to top of page, so that they can be accessed without needing to scroll through the entire type form
- moved versions list above type form display, to be grouped with other basic record information
- streamlined displayed basic record information ... only display UUID if needed, combine name and ID displays as appropriate for certain component types

Adding information about replaced wires to Tension Measurements action information pages
- tension measurements action now retrieves the number of replaced wires (per side) entered into the WINDING action for the same APA and layer
- number of replaced wires from winding is displayed alongside the number of re-tensioned wires determined by the DB

### DIFF
--- a/app/lib/Components_ExecSummary.js
+++ b/app/lib/Components_ExecSummary.js
@@ -373,6 +373,7 @@ async function collateInfo(componentUUID) {
 
       for (let i = 0; i < results[0].replacedWires.length; i++) {
         let singleWire_solderPads = results[0].replacedWires[i].solderPad;
+
         if (typeof singleWire_solderPads === 'number') {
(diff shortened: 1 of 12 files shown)


bjrebel (approved):
All changes look good

PR #226, opened 2024-06-28
Possible bug fix for Exec Summary crash ...

krishmaj:
- changed so that wire NCRs are only passed to the Exec. Summary if they actually contain information
- forgot that Formio automatically sets up a list containing a single, empty entry if a datagrid is part of the type form but is not used
- not 100% if this fixes the bug on Staging, since I can't replicate the crash on Dev ... but this change at least stops empty NCRs from being displayed


bjrebel (approved):
Good to catch the bug on staging

PR #229, opened 2024-07-02
Bug Fix for missing changed tensions when only on side B

krishmaj:
Lewis and Carlos found that the DB was not showing the expected list of re-tensionings on certain APA/layer combinations, even though the raw data showed that the values had been changed.

Found this mistake in the code, which would only occur if the re-tensioned wires were ONLY on side B.


bjrebel (approved):
Simple fix, good to go

PR #230, opened 2024-07-02
Bug Fix for missing changed tensions when only on side B

krishmaj:


bjrebel (approved):
tested on staging, good to go

PR #231, opened 2024-07-10
Changes to Determination of Re-tensioned Wires

krishmaj:
- the list of re-tensioned wires is now determined at the route level ... removed code from library function
- comparison is done between the two versions of the action that were most recently uploaded via M2M, accounting for which version of the action is being viewed
- the list is now passed to the Single Layer Tension Measurements action information page, instead of being written into the tension measurements action record
- new button on action information page, allows users to download a text file containing the list of changed tension measurements (both sides, and only if a comparison has been made) ... the text file's name shows the APA number and wire layer, as well as which versions were compared to get the list of changed tensions

Bug fix for incorrect x-axis labels on tensions bar plot
- code was previously first binning the data based on the minimum and maximum tension values, and then redrawing the x-axis based on fixed limits without rebinning the data
- code now uses the fixed limits when binning the data

Small updates to homepage PUG file
- changed page title that shows up in browser history to something a little less generic
- direct users to contact the DB Admin Team by email in case of problems, instead of the Github Issues page ... people already use the former anyway, never the latter
- added information about the weekly backup of production records to staging
- updated link to staging, and cleaned up some other text for clarity

### DIFF
--- a/app/lib/Actions.js
+++ b/app/lib/Actions.js
@@ -76,52 +76,6 @@ async function save(input, req) {
   newRecord.validity = commonSchema.validity(oldRecord);
   newRecord.validity.ancestor_id = input._id;
 
-  // If a tension measurements record is being EDITED, i.e. the new record's type form ID matches that of a tensions measurement action and there is an existing record ...
-  if ((newRecord.typeFormId === 'x_tension_testing') && (oldRecord)) {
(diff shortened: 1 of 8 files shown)


bjrebel (approved):
Looks ready to deploy to staging

PR #232, opened 2024-07-16
Improvements to interface level workflow display code ...

krishmaj:
- workflow steps are now unlocked only when the previous action has both been performed AND completed (for actions after free-form section, only when all free-form actions have been performed AND completed)
- tweaked displayed button and message text to clarify action performance vs. action completion
- changed designation of free-form section of APA Assembly workflow to account for new, non-free-form first action
- actions will now show as 'incomplete' if they have not yet been performed (previously showed as 'n.a.')
- general tweaks and improvements to interface code for workflow information pages ... clearer logic and separation of code depending on the step number and workflow type
- added Olga and Denis to list of personnel approved to sign off on Installation Survey results ... these two are a safe bet, and others can be added later

### DIFF
--- a/app/lib/utils.js
+++ b/app/lib/utils.js
@@ -105,5 +105,7 @@ module.exports = {
 
   // A dictionary of personnel who are authorised to sign-off on APA frame installation survey results
   dictionary_frameInstallationSignoff: {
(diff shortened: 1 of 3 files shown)


bjrebel (approved):
Looks fine to me.

PR #236, opened 2024-07-24
Tweaks and fixes to Frame Survey action interface pages ...

krishmaj:
- fixed a bug that would occur when both the actual and tolerance values in any row of either planarity results table are dashes (previously only accounted for the situation where one or the other was a dash)
- all calculated values now must satisfy the range given by: [-tolerance <= value <= +tolerance] to be acceptable (previously used < instead of <=)
- CERN-specific planarity results (LSS and Rib 1, etc.) are now hidden in the interface when displaying intake surveys ... they are still shown for the installation surveys, and the values still exist in the action records

### DIFF
--- a/app/pug/action.pug
+++ b/app/pug/action.pug
@@ -214,7 +214,7 @@ block content
             - outOfSpecCount = 0
 
             for [key, value] of Object.entries(action.data.intake_xCorners)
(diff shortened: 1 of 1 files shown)


bjrebel (approved):

PR #237, opened 2024-07-24
More tweaks to Intake Surveys action information page ...

krishmaj:
- pre-emptive code changes to account for some parameter tolerances being set to '-', indicating that these parameters are not used for frame acceptance (but should still be displayed)
- above code changes are relevant for both the manual cross corner measurements and the intake survey planarity results
- changed M2M frame survey results upload script to set certain tolerances to unused, following discussion with Ged and Eric
- changed Dashboard interface page title to match that of the default home page for consistency


bjrebel (approved):
reviewed and tested on staging

PR #239, opened 2024-07-28
Miscellaneous Requested Changes and Tweaks

krishmaj:
- re-instating code to disable warnings when uploading tension measurements (accidentally removed it in a previous commit)
- changing default user parameters in frame survey M2M upload script, to make it easier for myself in the future
- buttons in the 'Select Action Type to Perform' section of any component information page will now always be displayed in fixed alphabetical order, instead of changing order every time the page is refreshed
- setting the intake and installation surveys to be a free-form section within the Frame Assembly workflows
- small change to one of the search library functions' names, to make it consistent with the search's URL and other routing (same as all other search functions)

Extending Search for Action by Referenced Component
- added new option on interface page to search for winding actions (intended for finding referenced wire bobbin components)
- corresponding change in API route, and new bespoke library function to find all winding type actions that reference the provided UUID in the expected field
- removed unnecessary code from function for searching for board installation actions by referenced board UUID ... these actions are only ever performed on APAs, so there is no need to check the format of the component name - it will always be DUNE PID like

Removing ability to create new workflow components or perform new workflow actions outside of workflows
- components that are the subject of workflows (currently Assembled APAs and APA Frames) can no longer be created from the component information, single type component listing or component type listing pages
- actions that are part of workflows can no longer be performed from the component information, action information, single type action listing or action type listing pages
- non-workflow actions are still available to perform directly from a component's information page if applicable to the component type

### DIFF
--- a/app/lib/Forms.js
+++ b/app/lib/Forms.js
@@ -122,6 +122,9 @@ async function list(collection) {
 
   aggregation_stages.push(grouping);
 
+  // Sort the records into alphabetical order
+  aggregation_stages.push({ $sort: { formName: 1 } });
+
   // Query the specified records collection using the aggregation stages defined above
   let records = await db.collection(collection)
     .aggregate(aggregation_stages)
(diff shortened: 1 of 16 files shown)


bjrebel (approved):
changes look fine to me

PR #241, opened 2024-07-31
Implementing Tool for Comparing Tensions Across Locations

krishmaj:
This tool performs a comparison between tension measurements taken on the same layer of the same APA at two different locations, and displays the differences in the DB interface.  It replaces the 'Tension Measurements by UUID' search, which wasn't actually useful anyway.

Changes also include a reworking of the existing 'NumberArray' related Formio components - there are now 4 generic versions (displaying no plots, only a scatter plot, only a bar plot, and both plots), and a 'both plots' version that is specifically for displaying abolute or comparison tension measurements.


bjrebel (approved):
Changes look reasonable

PR #244, opened 2024-08-06
Changes to APA Executive Summaries ...

krishmaj:
- as requested by Eric, completely reworked the QC Signoffs section to match with newly implemented signoff actions within APA Assembly workflows
- changes made in library function and interface page .pug and .js files
- improved general layout of printable version of the summary ... moved 'pagebreak' .css to be commonly accessible (was previously only available in general component summaries)


bjrebel (approved):
looks good to test on staging

PR #245, opened 2024-08-06
Fixes for Problems in Exec. Summaries ...

krishmaj:
- added QA Check action IDs to interface console output, which will make it easier to track down issues in the future
- fixed incorrect loop structure that was preventing information from actually being retrieved (for...in --> for...of - the former loops through the INDICES, the latter loops through the ENTRIES)
- made some comments clearer as to which actions contain the various pieces of QA information


bjrebel (approved):
This is why we test on staging

PR #246, opened 2024-08-07
Additional APA/JSON action routes for use by M2M scripts

krishmaj:


bjrebel (dismissed):
Changes look fine to me.

PR #247, opened 2024-08-08
Bug fix for new API action route ... 

krishmaj:
... specified component UUID is a request PARAMETER, not a request query


bjrebel (approved):
Pretty simple change

PR #251, opened 2024-08-13
Attempting to implement in-situ check for component usability ...

krishmaj:
- when the user enters a component UUID into a Formio UUID box being used as part of a type form, the message can now display information about whether the selected component is ready for use or not
- currently, only applies to APA Frames - a frame is deemed usable if it has had a 'Final Frame QA Checklist' action performed on it (which can only be done via its assembly workflow), and the checklist action has been completed
- this does not stop the user from continuing with using an 'unusable' frame - it only informs them of the situation
- can be extended to any other type of component, as long as there is a way to define its usability using existing DB records and routes


bjrebel (approved):
Looks like a good feature to implement

PR #252, opened 2024-08-19
Miscellaneous Updates and Tweaks

krishmaj:
Small updates to M2M scripts and README
- all backend M2M functions now use the 'json.loads' function to deserialise the server response ... previously some were using explicit string manipulations instead, which does work but isn't really the correct way to do it
- increased the M2M connection timeout to 60 seconds ... found that some retrieval requests relating to workflows take more than the previously used 10 seconds
- added examples for the recently added GetWorkflow and GetListOfWorkflows backend functions to the GetRecords template script
- updated M2M and general READMEs with some requested changes for clarity

Additional component usability checks (requested by Sotiris)
- grounding mesh panels should have had a QA Inspection performed, and the inspection disposition should be 'use as is'
- wire bobbins should have a mean break strength >= 24 N

Changes to determination of workflow status
- status is now given as the percentage of all action steps that have been completed, instead of just a binary complete vs. incomplete string
- when displayed on the workflow information and listing pages, the completion percentage is colour-coded for convenience

Splitting listings of UK vs. US APA Assembly workflows (requested by Sotiris)
- the workflow type listing page now has separate entries for UK and US APA Assembly workflows
- each of these entries now links to a list of ONLY the APA Assembly workflows that have been performed in the corresponding location
- location matching is based on the DUNE PID of the corresponding APA component ... workflows that do not have an associated APA are by default put into the UK list
- overall all-types workflow listing page and listing pages for other workflow types are unchanged

Small change to component information page layout (requested by Brian)
- the 'create new component of this type' button is now positioned below the others, and with a space between them to avoid accidental usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,19 +31,19 @@ docker compose down
 The code in this respository is organised as follows (for simplicity, normally hidden directories and files are not listed here):
 
 * `/app` : contains all of the code required for the APA DB app
-    * `/lib` : JavaScript functions that operate directly on and with the MongoDB database
-    * `/pug` : Pug templates for the web interface pages
-    * `/routes` : JavaScript functions that dictate which `/lib` functions are called when a user accesses a specific web interface URL, and which `/pug` template is displayed for that URL
+    * `/lib` : server-side JavaScript functions that operate directly on and with the MongoDB database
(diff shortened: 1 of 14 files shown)


bjrebel (approved):
Changes look reasonable

PR #253, opened 2024-08-20
Follow-up tweak and bug fix ...

krishmaj:
- grounding mesh panels should additionally be set as 'ready to use' if their QA disposition is 'conformant'
- bug fix for the API workflow listing route ... this one does not use the 'options' parameter (unlike the equivalent interface route), so it was failing to find a non-existent 'location' field


bjrebel (approved):
Looks ready to test in staging

PR #256, opened 2024-08-21
Bug fix - forgot to use the actual type form ID for links, instead of the location-modified one

krishmaj:
Confirmed to work correctly on Staging


bjrebel (approved):
checked on staging

PR #257, opened 2024-08-23
New search for APAs that have reached a certain assembly workflow step

krishmaj:
- new library, API and interface routes
- additions to 'Search for APAs by Production Details' interface .pug and JS code
- updated sidebar links and search descriptions page

Will need to be tested on Staging


bjrebel (approved):
Changes look good to me. Time to test on staging

PR #258, opened 2024-09-17
Miscellaneous Requested Changes

krishmaj:
Updating default file locations in Frame Surveys M2M upload script

Changes relating to Geometry Board Rejection at Factories
- added new optional parameter to 'Search for Geometry Boards by Location or Part Number' to allow filtering by board status - accepted or rejected
- changes to API routes and library functions
- changes to interface page .pug and .js codes ... reworked layout, cleaned up static code, and search must now be explicitly started using the interface button
- updated Search Descriptions page
- when referencing a geometry board using the Formio Component UUID entity, the info message will now indicate if the board has been accepted or rejected

Changing action history display on information pages of workflow-related components
- for workflow-related component types (Assembled APAs and APA Frames), the 'action history' section of the component information page will now only show non-workflow actions
- a message is also now shown, directing users to view the component's workflow for accessing workflow actions (and the workflow is hyperlinked)
- non-workflow components are unchanged - all actions are shown in the action history section
- added hyperlink to workflow information page in the existing message that is displayed under the 'Select Action to Perform' section for workflow-related components
- made the section titles more explicit regarding workflow vs. non-workflow actions for workflow-related components

Requested tweak to Single Layer Tension Measurement action information pages
- the number of changed wires found in the corresponding winding action is now always shown, even if no inter-version tension comparison is available

Additional security checks when performing or editing certain action types
- it has been reported that certain people have been submitting empty APA QA Check actions on behalf of factory leads, in order to unlock APA Assembly workflow steps ahead of the current progress
- changed DB action submission function so that it now checks if the submitting user (i.e. the person who is logged in) is actually one of the factory leads (or DB admins) ... if not, the user will not be permitted to submit the action, and a message will be shown
- check is done on the user's Auth0 account ID, which cannot be changed ... better than the user's display name
- currently only applicable to 'Assembled APA QA Checks' actions ... all other actions can still be submitted by anyone with the correct basic permissions ... but this check can be extended to other action types if needed

Additional checks on user-submitted data when performing or editing certain action types
- similarly to previous commit, it has been reported that certain people have been submitting empty but 'completed' actions, in order to unlock APA Assembly workflow steps ahead of the current progress
- changed DB action submission function so that it now checks for the presence of actual data in certain fields if the action is set to 'complete' (specific to particular action types) ... if there is no data, the submission is denied and a message will be shown
- currently this only applies to 'Single Layer Tension Measurement' actions that have no tension measurements, but it can be extended to other action type / field combinations

### DIFF
--- a/app/lib/Actions.js
+++ b/app/lib/Actions.js
@@ -7,6 +7,7 @@ const { db } = require('./db');
 const dbLock = require('./dbLock');
 const Forms = require('./Forms');
 const permissions = require('./permissions');
+const utils = require('./utils');
 
 
 /// Save a new or edited action record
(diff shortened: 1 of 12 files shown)


bjrebel (approved):
Changes look reasonable for testing on staging

PR #260, opened 2024-09-18
Changes to the Search for Geometry Boards by Location or Part Number ...

krishmaj:
- added new optional parameter for checking if a tooth strip attachment action has been performed or not ... this can be selected in addition to the existing board acceptance status optional parameter
- corresponding changes to API routes and library function (rewrote the post-MongoDB parts of the latter for better logic when working with two optional parameters that can each take one of three values)
- fixed a bug where the displayed UKIDs in the search results did not always match the actual board that was being hyperlinked to (the results themselves were always correct - just the displayed UKIDs were sometimes wrong)
- updated Search Descriptions page
- [unrelated] small tweak to the newly added text for displaying how many changed wires were found in a particular layer's winding action from the tension measurement action

### DIFF
--- a/app/lib/Search_GeoBoards.js
+++ b/app/lib/Search_GeoBoards.js
@@ -6,7 +6,7 @@ const { db } = require('./db');
 
 
 /// Retrieve a list of geometry boards that are at a specified location across all part numbers
-async function boardsByLocation(location, boardStatus) {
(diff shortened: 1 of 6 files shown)


bjrebel (approved):
Previously tested on staging

PR #264, opened 2024-09-23
Reworking logic for searches involving user-changeable parameters ...

krishmaj:
- the current search logic (shared by all searches) fails if one or more of the required parameters can be changed by users ... the search will return matching results for both the original and user-changed versions of the associated record
- this is because the aggregation stage for matching to the parameters is done before the stage for selecting only the latest version of the record
- attempting to fix this by splitting the matching stage into multiple stages ... the first one (matching to component or action type form ID and/or component UUID or action ID) will remain as is - before version selection ... a second matching stage will now be done after the version selection, and will perform the matching to any user-changeable parameters
- new logic has been applied to the 'Search for Geometry Boards by Order Number', since this was the one where the issue has been specifically identified, plus the 'Search for Geometry Boards by Visual Inspection Disposition', since it shares the same interface page as the former
- if it works, the same changes can be applied to all other searches that involve matching to user-changeable parameters

### DIFF
--- a/app/lib/Search_GeoBoards.js
+++ b/app/lib/Search_GeoBoards.js
@@ -9,7 +9,7 @@ const { db } = require('./db');
 async function boardsByLocation(location, acceptanceStatus, toothStripStatus) {
   let aggregation_stages = [];
 
-  // Retrieve all 'Geometry Board' records that have the same location as the specified one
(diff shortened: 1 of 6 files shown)


bjrebel (approved):
Looks good to try out on staging

PR #265, opened 2024-09-25
Major Updates to All Searches

krishmaj:
Backend Libraries
- updated various search functions with revised logic to handle the situation where the matching parameter(s) can be edited by users
- the fix involves splitting the 'match' aggregation stage into two ... the first one is performed before version selection and matches against fixed record fields such as component UUID, type, etc., and the second one is performed after version selection and matches against any remaining search parameters that are not necessarily fixed between versions
- fixed issues with mismatched component UUIDs and displayed DUNE PIDs when searching for grounding mesh panels by location or part number (the same issue was previously fixed for geometry board UUIDs and UKIDs)
- pre-emptively added code for handling component location of 'installed on APA' when searching for populated board kit components by location
- general code cleanup and consistency changes, reduced unnecessarily long comments

API Routes
- changed all (but one) routes to use request parameters instead of request queries ... makes it easier to code in the backend libraries
- the one remaining route that uses request queries is the search for board shipments by reception details ... this is already complicated, but works fine, so no need to mess with it

Interface Pages (.pug and .js files)
- all searches that involve anything more than a single UUID or ID input must now be initiated using the appropriate button on the interface page ... this allows more control over the query as it goes to the server, and stops multiple queries from being sent simultaneously
- made the interface pages more consistent in layout and formatting
- general code cleanup and consistency changes in the .js files, reduced unnecessarily long comments
- updated search descriptions page for all of the above changes


bjrebel (approved):
Seems like sensible changes

PR #268, opened 2024-10-06
Small formatting change to fix button that is slightly too wide

krishmaj:
Confirmed to fix the problem on Staging


bjrebel (approved):
Checked on staging

PR #269, opened 2024-10-10
Requested Changes to Returned Geometry Boards and Searches

krishmaj:
Displaying more component information for Returned Geometry Board Batches
- component info page for Returned Geometry Board Batches now shows a table of board UUIDs and UKIDs, similar to existing setup for board and mesh shipments and batches
- QR code links are also displayed because they come for free as part of the table structure
- type form on component info page no longer shows the integrated table of UUIDs for Returned Goemetry Board Batches, since it is no longer needed
- additional section in corresponding component info page route to collate the board information into the standardised format for table display

Improvements to displayed results for various searches
- when search results are grouped by a certain parameter, the results within each group are now fixed in order (previously would be displayed in a random order)
- for results specifically involving geometry boards, ordering is by decreasing UKID ... other components and actions ordered by some other reasonable parameter (typeRecordNumber, action ID, etc.)
- ordering is done using MongoDB sort functionality as part of aggregation stages, so there shouldn't be much increase in time taken for the search (and this is the fastest way anyway, so even if it is slower there's nothing we can do about it)

Removing specialised user scripts from M2M directory
- the existing specialised user scripts for uploading tension measurements and frame survey results have been moved to other Github repositories, so they don't need to be here any more

Small changes requested by Sotiris
- component listing pages now show bobbin ID for wire bobbins, in place of UUID
- increased limit of number of components to retrieve to 500 for API single-type component listing route (only used by external M2M script, interface route limit remains at 200)


app/lib/Search_ActionsWorkflows.js
bjrebel, line 1: Why did you repeat the function code? Can't it just be abstracted out once and then called in multiple places?

bjrebel (commented):
One question about duplicated code, will wait for reply before approving

app/lib/Search_ActionsWorkflows.js
krishmaj: Thanks for pointing that out - fixed the duplication and tested to confirm.

bjrebel (approved):
Looks good, let's test on staging

PR #270, opened 2024-10-14
Fixing incorrect variable names left from previous commit ...

krishmaj:
... plus correcting descriptions of search result ordering


bjrebel (approved):
Glad that staging was useful in catching this bug

PR #272, opened 2024-10-17
Changes to Post-Submission Behaviour for Components and Actions

krishmaj:

The general idea is to remove unnecessary internal redirections, and instead perform all post-submission behaviour as part of the actual submission process.  (The post-submission behaviour in question is all about updating locations and dates relating to installation or reception, depending on the type of component or action being submitted.)

This PR also removes the on-the-fly calculation of workflow completion status, which was severely slowing down the workflow listing pages, even with just 20-ish workflows present.  This has been replaced with calculation of the completion status being done whenever a workflow action is submitted, and only for the specific workflow in question ... the status is now also saved directly into the workflow record, and can be accessed directly from there without any need for calculation.


bjrebel (approved):
Very reasonable improvements

PR #274, opened 2024-10-19
Another attempt at the bug fix for displaying completion status ...

krishmaj:
- at the moment, the completion statuses will mostly be 'undefined' ... and as per the previous attempt at this fix, if the status is exactly zero, JS interprets it as 'null'
- using the '!== null' comparison only checks if it is 'null', not 'undefined' ... but since they are the same type of value, using '!= null' catches both

### DIFF
--- a/app/pug/workflow.pug
+++ b/app/pug/workflow.pug
@@ -33,7 +33,7 @@ block content
 
             dt Completion [%]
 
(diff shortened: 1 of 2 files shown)


bjrebel (approved):
This is why we deploy staging first

PR #276, opened 2024-10-23
Testing new development setup with some small commits

krishmaj:
- added Gwenn to the list of personnel who are allowed to sign off on frame intake
- removed 'version' attribute from docker-compose file ... apparently this has been deprecated for a while now, and no longer does anything.  And it was creating a warning message on the new dev setup


bjrebel (approved):
Completely uncontroversial change

PR #278, opened 2024-10-26
Miscellaneous Tweaks and Fixes

krishmaj:
Partial fix for incorrectly ordered workflow actions in Assembled APA component summaries (other part of the fix requires changes to the actual records)

Fixes and tweaks to the displayed results for APA Frame Installation Survey actions ... some frames do not have certain Envelope Analysis parameters (due to limited data), but the current code is not able to display the table of results correctly if any parameters are missing

Bug fix for the workflow route being used by the STFC M2M APA Information script ... forgot to previously update it to use the new workflow completion status setup - was still using the old, manual calculation instead of just extracting the value from the workflow record

Fix to the displayed results for APA Frame Intake Survey actions ... some frames do not have all 3 cross corner measurements, and the missing ones should be indicated by '-' to keep them consistent with other missing survey parameters.  However, the code is currently not able to display such missing values

### DIFF
--- a/app/lib/Actions.js
+++ b/app/lib/Actions.js
@@ -257,6 +257,7 @@ async function list(match_condition, options) {
       typeFormId: true,
       typeFormName: true,
       componentUuid: true,
+      workflowId: true,
       validity: true,
     }
   })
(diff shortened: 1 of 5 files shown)


bjrebel (approved):
Let's get these going on production

PR #280, opened 2024-10-29
Bug fix for being unable to edit non-workflow actions ...

krishmaj:
- for non-workflow actions, the 'workflowId' object doesn't exist ... and is therefore passed as 'undefined' (NOT 'null')
- crash was being caused by the route code only checking against 'null', not both 'null' and 'undefined', and therefore passing nothing to the interface
- fixed by relaxing the type equality, meaning that both 'null' and 'undefined' are compared (since they are the same 'type' with different values)
- also removed checks for source maps in local libraries ... these are not necessary since they are local, but these files were originally external libraries (for which source map checks are preferable)


bjrebel (approved):
Try it out

PR #284, opened 2024-11-04
Adding new dictionaries of UW personnel as requested

krishmaj:
Also made the previously added API route to the list of Manchester geometry board metrology technicians CamelCase, to keep it consistent with the other routes (will need to change the type form on Production to match).

Also added some console output to the Formio Component UUID entity ... might be useful for debugging problems with the QR code scanning and decoding part of the DB interface.  Is permanent, but will only show up with 'persistent logs' enabled in the debug console, so it shouldn't get in anyone's way.


bjrebel (approved):
All look good to me

PR #286, opened 2024-11-14
New Search for Components by DUNE PID

krishmaj:
QR code URL is now displayed on all component information pages

New Search for Components by DUNE PID
- renamed previous 'Search for Components by UUID or Type and Record Number' to 'Search for Components by Identifier' in interface pages and route
- added new input box for DUNE PID, and corresponding interface page JS code for retrieving the input string and passing it through the (new) route
- added route and library function for new search, and updated Search Descriptions page


bjrebel (approved):
Let's give these changes a go on staging

PR #288, opened 2024-11-19
Requested Changes from 2024-11-19 Meeting

krishmaj:
- adding new personnel list for APA frame NCR concessions signoff
- adding general sorting functions to 'utils' library - these are exactly the same as the sorting functions that were previously in the various search-related libraries, but now they are centralised to avoid code duplication
- changed various libraries to use the new general sorting functions in 'utils' library
- added new APA NCR types to dictionary of types in library function (frame and mesh issues)

- removed APA frame NCRs from executive summaries ... we have decided not to put frame NCRs into the APA DB, so no attempt needs to be made to retrieve and display them
- fixed apparent bug in executive summaries library function that was stopping non-wire APA NCRs from being retrieved ... surprised that this wasn't spotted sooner

- removed 'APA Frame' option from Search for NCRs by Component Type ... same reason as for removing APA frame NCRs from executive summaries
- added NC type to the displayed results when performing a Search for NCRs by either Component Type or UUID, and adjusted layout and formatting accordingly

- added new library function to search for board shipments that reference a specific geometry board by UUID ... this is NOT a new search - it will be used to retrieve more detail about a geometry board's history, but it fits best alongside the various component-related searches

### DIFF
--- a/app/lib/Components_ExecSummary.js
+++ b/app/lib/Components_ExecSummary.js
@@ -1,10 +1,8 @@
 const MUUID = require('uuid-mongodb');
 
-const Actions = require('./Actions');
 const Components = require('./Components');
 const { db } = require('./db');
 const Search_ActionsWorkflows = require('./Search_ActionsWorkflows');
-const Workflows = require('./Workflows');
 const utils = require('./utils');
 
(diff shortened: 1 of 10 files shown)


bjrebel (approved):
Changes look good, ready to test

PR #289, opened 2024-11-25
New action history setup for geometry boards ...

krishmaj:
- added a new 'action history' setup for geometry boards ... a new, more detailed table of actions that have been performed on each board is now shown on the component info pages
- new table replaces generic history list (only for geometry boards), and also contains information about any shipments that the board was part of


bjrebel (approved):
Changes look good to me

PR #290, opened 2024-11-30
Fixing bug with NCR search by component type ...

krishmaj:
- currently, the search matches against ANY version of the NCR record, but only returns the latest one ... even if the latest one wasn't the one that matched
- fixed so that the search now matches only against the latest version of the record - the match to component type is now performed after the latest version has been retrieved (alongside the already existing matches to disposition and status)
- now consistent with other searches that match against user-editable record fields


bjrebel (approved):
Looks good to me

PR #294, opened 2024-12-17
New backend code for geometry board counts ...

krishmaj:
- new library function and API route for retrieving the counts of geometry boards at each [board part number, board location] combination
- new library function and API route for retrieving the counts of geometry board rejections at each [board part number, rejection location] combination
- renamed library function and related route and interface code variables for the existing counts per component type functionality, to make things clearer alongside new counting functions
- added comments to existing counts per component type library function

These changes relate to a new M2M script being written for use by Andra/Sandeep for tracking board intake and movement, and comparison to orders for payment, etc.

### DIFF
--- a/app/lib/Actions.js
+++ b/app/lib/Actions.js
@@ -322,6 +322,84 @@ async function list(match_condition, options) {
 }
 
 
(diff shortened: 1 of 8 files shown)


bjrebel (approved):
Let's give it a go

PR #295, opened 2024-12-18
Follow Up Fixes and Changes related to Previous Commits

krishmaj:
Bug fixes in searches for geometry boards by part number or location
- the current code for searching by part number matches to all versions of a board record, not only the most recent
- it was written as such because I wasn't aware that the part number can change between record versions (it shouldn't, but it can and does)
- fixed search such that the match to part number is now done AFTER selecting only the most recent version of each board record
- made the equivalent change in the search by location ... again, the location shouldn't change between record versions, but who knows how things might behave in the future

Additional changes to library functions associated with previous bug fix
- changed 'Count Board Rejection Actions' library function so that disposition matching is done after retrieving most recent version of each action
- changed logic of 'Count Boards' library function ... the previous code flow was not working as intended (was an attempt to do things more efficiently), so it has been changed to something more equivalent to existing functions that do similar things

### DIFF
--- a/app/lib/Actions.js
+++ b/app/lib/Actions.js
@@ -327,11 +327,10 @@ async function list(match_condition, options) {
 async function boardRejectionCounts_byPartNumberAndLocation() {
   let aggregation_stages = [];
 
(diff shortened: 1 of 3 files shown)


bjrebel (approved):
Let's test these out

PR #297, opened 2024-12-24
Request, Miscellaneous Improvements and Cleanup

krishmaj:
Miscellaneous improvements ...
- the list of part numbers in the 'Search for Geometry Boards by Part Numbers' is now shown in numerical part number order, instead of alphabetical description order
- added more clarity to search descriptions as to how geometry board and grounding mesh panel locations are set
- added the ability to set a 'null' location when searching for geometry boards by location ... this will allow easier debugging of the missing locations (NOTE: this is not available through the interface - it is a 'hidden' option when using a direct URL to the search results)

Requested Functionality: Viewing QR Codes in Bulk ...
- new functionality for viewing component QR codes in bulk, for a sequential range of type record numbers within a single component type
- new interface page for selecting the component type form ID and type record number range ... upon confirmation, this page redirects to the already existing page for viewing and printing the QR codes (currently used for viewing all QR codes of components in a batch)
- corresponding new interface and mid-level routes ... no new library function necessary, since collation of the short UUIDs can be done using the existing function to retrieve a single component by specifying its type form ID and type record number
- added new link in 'Components' section of the sidebar ... this isn't really a search, so it doesn't belong there ... it's more like a tool - akin to the existing 'Compare Tensions' action tool
- renamed existing .pug file for viewing multiple QR codes, to reflect the additional new functionality

Cleaning out unused code ...
- removed 'jest' code for testing routes ... we don't use it at all, and I'm fed up of trying to keep it up-to-date for no reason
- removed interface code for viewing type form tags ... nobody uses tags apart from the 'Trash' one (so the 'tags' framework still has to remain for that)
- removed interface page for viewing tags, corresponding route to interface page, and corresponding library function to collate tags to be passed to interface page


bjrebel (approved):
give it a go

PR #298, opened 2024-12-30
Small fix for missing component names ...

krishmaj:
- if a board record has an empty 'data.name' field (i.e. the field exists, but is an empty string rather than the DUNE PID), the component UUID will now be shown instead
- currently, if the field is empty the entire component listing page just crashes
- note that the field shouldn't ever be empty in the first place, and if so it's a sign that something went wrong during submission ... but at least now users can still carry on using the DB while admin looks into the issue


bjrebel (approved):
Good to create this fail-safe

PR #299, opened 2025-01-01
More bug fixes ...

krishmaj:
- additional fix for missing component names ... now accounting for if the 'data.name' field is completely missing
- attempting fix for missing reception information when submitting components via the M2M system ... when doing so, the client-side URL that is supposed to assign sub-component location and date parameters is bypassed, and so these fields end up being 'null' - this change moves that assignment to the server-side library function, so it should be done regardless of whether the component is submitted via the interface or via M2M
- added some comments in the Formio ComponentUUID entity code to explain a certain console error that can sometimes show up ... no actual fix needed, just an explanation for future reference


bjrebel (approved):

PR #300, opened 2025-01-15
Requested new M2M functionality for performing board QA actions ...

krishmaj:
- new API route to retrieve a component record via its type form ID and type record number (the library function associated with this functionality already exists)
- small changes to variable names in existing library and search functions to make things more explicit and consistent
- new M2M backend function to retrieve a component record via its type form ID and type record number, accessing the new API route, plus small change to existing 'GetComponent' function to make the difference between that and the new one more explicit
- updated M2M README with information about the new backend function


bjrebel (approved):
Let's give it a go

PR #302, opened 2025-01-22
Adding requested users to personnel list ...

krishmaj:
Adding only Kim Williams to the list of factory technicians.

The original changes (adding Dave Brown, Gwenn Mouster and Dan Salisbury to the list of personnel who are authorised to sign off on APA QA Checks) have been reverted, since we're waiting for a decision on that, but Kim needs to go in ASAP.


bjrebel (approved):
Add a user, nothing to really check here.

PR #306, opened 2025-02-07
Displaying Installed Geometry Boards and Guarding Against Retroactive Reception

krishmaj:
Adding display of installed geometry boards to APA component info ...
- the component information pages of Assembled APAs now show a table of all installed geometry boards (UKID, part number and string, link to board info page)
- added new library function, modified existing component info retrieval route, modified existing component info page .pug file

Guarding against retroactive board shipment reception ...
- the reception information of individual geometry boards will no longer be updated if they've already been installed on an APA (reception.location = 'installed_on_APA')
- this should prevent boards from being 'reset' to their last shipment reception location if the shipment is retroactively received somewhere

### DIFF
--- a/app/lib/Components.js
+++ b/app/lib/Components.js
@@ -165,36 +165,47 @@ async function save(input, req) {
 }
 
 
-/// Update the most recently logged reception location and date of a single component
+/// Update the most recently logged reception information of a single component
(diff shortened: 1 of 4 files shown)


bjrebel (approved):
Looks good to test on staging

PR #310, opened 2025-02-26
Consolidating Searches by Location and Part Number

krishmaj:
Consolidating component searches by location and part number ...
- currently we have 3 separate 'Search for XXX by Location' component searches (and more have been requested), and 2 separate 'Search for XXX by Part Number)
- the code for each type of search does not vary greatly from component to component, so they can be consolidated into a single search algorithm for each type
- new interface pages and associated JS files - users must now select the component type and location or part number (plus optional parameters for geometry boards)
- new associated interface routes, API routes, library functions and new sidebar links
- existing search code has been removed completely, and search descriptions have been updated

Changing endpoint of APA Assembly workflow freeform section ...
- freeform section now ends with step 12 (X Layer Foot Board Installation), as this has been moved to before the Frame Prep QA Checks


bjrebel (approved):
Let's give it a go

PR #311, opened 2025-02-27
Fixing incorrect DUNEPIDs for cable harnesses ...

krishmaj:
- DUNEPID is different for side A vs. side B cable harnesses


bjrebel (approved):
That is what we needed.

PR #315, opened 2025-03-14
Changes relating to new Board Installation actions ...

krishmaj:
- submission of the new board installation actions will set the individual board locations to 'installed on APA' with corresponding date and APA UUID
- submission of the old board installation actions will NOT trigger update of the individual board locations any more
- searching for boards by installation action will now match against the new board installation action types ONLY, not the old ones
- reduced verbosity of text displayed under Formio Component UUID input boxes ... was unnecessarily long and messing up the formatting of certain type forms


bjrebel (approved):
Let's give it a go

PR #317, opened 2025-03-15
Fixing length of APA Assembly workflow freeform section ...

krishmaj:
... to account for new board installation actions


bjrebel (approved):
Simple change, looks good

PR #321, opened 2025-04-16
Change to APA Frame starting location ...

krishmaj:
- frame prep has recently been moved back to the Daresbury Factory (was previously being done at the UK Warehouse)
- changed starting location of APA Frames from 'ukWarehouse' to 'daresbury' accordingly
- list of actions performable on 'APA Shipment' type components now displays as collapsed by default, in line with other workflow-related component types


bjrebel (approved):
All look good

PR #323, opened 2025-04-17
Recent Changes from Staging

krishmaj:
Change to APA Frame starting location
- frame prep has recently been moved back to the Daresbury Factory (was previously being done at the UK Warehouse)
- changed starting location of APA Frames from 'ukWarehouse' to 'daresbury' accordingly
- list of actions performable on 'APA Shipment' type components now displays as collapsed by default, in line with other workflow-related component types

Bug fix for Geometry Boards internal server error
- error was being caused by the 'toISOString()' function on the component info page being applied to a non-Date type variable
- the 'lastEditDate' variable in the first (component creation) 'action' in the Geometry Board history was sometimes being passed as a string, and sometimes as a Date ... no idea why it was inconsistent
- fix is to explicitly cast it to a Date on the server side, regardless of whether it is one already or not

### DIFF
--- a/app/lib/Components.js
+++ b/app/lib/Components.js
@@ -117,8 +117,8 @@ async function save(input, req) {
     newRecord.reception.detail = '';
 
     // Components of certain types will always start at specific fixed locations, whereas the rest do not need any initial location set (only for the record field to exist)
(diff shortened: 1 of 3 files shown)


bjrebel (approved):
Changes tested on staging

PR #324, opened 2025-05-05
Adding component names for APA Shipments ...

krishmaj:
- the component name for APA Shipment type components should show which APAs are in the shipment
- added new section in 'Components.save' library function for assigning the shipment name (this can in principle be done at any time, since the APAs are editable by users)
- changed existing comments about general component naming, in preparation for a big cleanup of this stuff in the near future


bjrebel (approved):
looks reasonable to me

PR #327, opened 2025-05-12
Moving to MongoDB version 7.0 (6.0 is reaching end of life)

krishmaj:
As required by Fermilab for the Staging and Production instances.
Have tested all functionality on Development - no issues found on the software side.  Fermilab-based deployments should also be tested for any build issues.

Also added a workaround for 'null' shipment reception
- if the component 'reception' object is 'null' currently the DB will crash in two places - when attempting to set a new location (if editing the component), and when attempting to display the location (in the case of geometry boards which were part of a shipment with 'null' reception)
- have changed the component information interface page so that if a shipment has a 'null' reception object, it will now display a message indicating as such, rather than breaking completely
- have changed the component submission library function so that if a component is being edited, but has been passed a 'null' reception object, the object will now be first reconstructed and populated with fields (before attempting to assign values to the fields)
- this doesn't actually answer the underlying question of why a component's reception object can be 'nulled' in the first place, but it should at least allow users to work around it

### DIFF
--- a/app/lib/Components.js
+++ b/app/lib/Components.js
@@ -139,7 +139,17 @@ async function save(input, req) {
       newRecord.reception.location = '';
     }
   } else {
-    newRecord.reception = input.reception;
+    // For some unknown reason, sometimes the entire 'reception' object can be set to 'null' when editing a component (seems to happen rarely, and only with 'Board Shipment' components) ...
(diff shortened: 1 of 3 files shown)


bjrebel (approved):
Looks good to me

PR #328, opened 2025-06-05
Overhaul of Component Names (Part 1) ...

krishmaj:
- component names and DUNE PIDs are now always assigned for ALL component types, with the name format being dependent on the type
- some component names are assigned once at component creation, while others are (re)assigned any time the component is edited (and at creation) - this is to account for any changes to user-definable fields that the name uses
- added new Utilities dictionary for mapping geometry board part numbers to their corresponding DUNE PID 'component ID's
- added new 'Administrator Utility' flow ... interface page and JS, interface and API routes - this is intended to be a generic flow that can be used for whatever admin-level utility is required ... code is written in such a way that it only needs a specific library function and interface redirection to be changed for a different utility
- new function for setting the component names and DUNE PIDs of all versions of all records for a specific component type (using essentially the same code as would be used to assign them in the first place)

### DIFF
--- a/app/lib/Components.js
+++ b/app/lib/Components.js
@@ -7,6 +7,7 @@ const { db } = require('./db');
 const dbLock = require('./dbLock');
 const Forms = require('./Forms');
 const permissions = require('./permissions');
+const utils = require('./utils');
 
 
 /// Generate a new component UUID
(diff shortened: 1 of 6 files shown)


bjrebel (approved):

PR #330, opened 2025-06-10
Fix on the Fix

krishmaj:
Forgot that the validity date is stored as an actual Date object, not a string


bjrebel (approved):
yup, looks like a good change

PR #331, opened 2025-06-11
Update buildconfig.yaml

bjrebel:
Update the location from which the node image is pulled


krishmaj (approved):
Seems to have fixed the build problem.

PR #332, opened 2025-06-11
Overhaul of Component Names (Part 1)

krishmaj:
Adding the 'Administrator Utility' to Production, so that the new 'data.componentName' and 'data.dunePid' fields can be added to every component record.  Tested on Staging.  Changes to the interface to actually use the new component names will follow in a separate PR.

Also some small updates to the OKD build and README files, and a workaround for the 'null' reception issue on board shipments (not yet tested because it's a sporadic issue that seems so far impossible to repeat on demand).


bjrebel (approved):
Looks good to me.

PR #333, opened 2025-06-11
Overhaul of Component Names (Part 2) ...

krishmaj:
- changes to various interface pages so that they actually display the new component names (and DUNE PIDs) as needed
- UUID input and other autocomplete boxes will now also make use of the new component names


bjrebel (approved):
Looks ready for testing on staging

PR #334, opened 2025-06-11
Fix for when the 'validity.startDate' is a string ...

krishmaj:
- for whatever reason, geometry board 19536 doesn't have a typeRecordNumber or UUID, but still exists as a 'GeometryBoard' type record ... with a string-type 'validity.startDate' (instead of being a 'Date')
- code will now check for the type ... if 'string', do not convert to ISO string before substringing (since a string cannot be converted) ... otherwise do as normal


bjrebel (approved):
Good catch

PR #335, opened 2025-06-11
Overhaul of Component Names (Part 2)

krishmaj:
Changes to various interface pages, so that they now make use of the new component names and DUNE PIDs.
Formio 'ComponentUUID' entity and autocomplete suggestions will now also display the names correctly and consistently.

Also includes the fix for when the 'validity.startDate' field is a string, instead of a 'Date' type object.  (Literally only relevant for one geometry board in the entire DB, but it breaks the administrator utility function, and is stopping the additions of new name fields in geometry boards, if not accounted for.)  Fix has been tested on Staging, confirmed to work.


bjrebel (approved):
Changes checked on staging

PR #338, opened 2025-06-19
Adding APA Shipments to Administrator Utility ...

krishmaj:
Overlooked this component type previously, since I don't have any of them on my Dev instance
Needs testing on Staging, since Dev is currently a bit broken due to other (unrelated changes)


bjrebel (approved):
Let's test on staging and see if it works as intended

PR #340, opened 2025-06-20
Adding new STFC technicians to interface list

krishmaj:
Have also made new DB accounts for Mitchell and Nigel ... Abbie already has one, and was just missing from the interface list for some reason.


bjrebel (approved):
Adding names is an easy approval

PR #344, opened 2025-06-29
Upgrading Node and all NPM dependencies ...

krishmaj:
Most of the dependencies have upgraded to the latest versions without needing any code changes ... please see below for details on those that did.

`express` 4.18.1 --> 5.1.0
- `res.json()` and `res.send()` must now first specify the status explicitly --> `res.status(201).json()` and `res.status(200).send()`
- the arguments for `res.redirect(url, status)` have been swapped around --> `res.redirect(status, url)`
- the `?` and `|` characters are no longer supported within route URLs ... the only place we were using these was in `/routes/api/forms.js` - I've reworked the offending route URLs to both remove the characters, and also remove some previous ambiguity
- regular expressions are no longer supported within route URLs ... we had been using them to ensure correct formatting of component UUIDs and action/workflow IDs, but this was in fact unnecessary at the route level anyway - they were already being checked and correctly formatted before being passed to the routes ... all regex's have been removed from routes and the `utils.js` library
- related to the above, removing the regex's resulted in ambiguity between routes previously ending with `/:typeFormId` and `/:id` ... I have changed the former to now be `/:typeFormId/new`
- for some reason, two of the 'GET' jQuery.ajax routes (`/json/administratorUtility` and `/json/search/boardShipmentsByReceptionDetails`) no longer work (returning nothing or an empty JS object), but switching them to 'POST' fixes them ... no idea why only these two are affected and none of the others are

`auth0` 2.44.0 --> 4.26.0
- various changes to API function calls and namespaces in `/app/routes/users.js`
- removed the 'list of human users' interface page and associated routes and library functions ... I couldn't get it to work correctly with the new `auth0` version, but I also realised that it was completely redundant - it was only ever accessible by DB admins, who also have access to far more information about users using the actual Auth0 website
- removed the associated, and now no longer used, `users:view` permission from all three Auth0 tenants

`mongodb` 4.10.0 --> 6.17.0
- changed all variables named `ObjectID` to `ObjectId` ... previously, these were two completely separate but entirely equivalent classes, but we actually mixed them up - calling the former and naming it the latter.  The former has now been removed from `mongodb`, so everything should now call and be named the latter.
- `db.collection()` must now return a promise --> `await db.collection()` ... most instances of this already did so, with just a couple needing changing
- `findOneAndX` type functions must now explicitly specify `includeResultMetadata: true` in order to return the `result.ok` parameter (which we need)
- the `useNewUrlParser` and `useUnifiedTopology` options in `/app/lib/db/mongodb.connection.js` no longer do anything and will be deprecated soon, so I have removed them now in preparation

`node-sass` and `express-compile-sass`
- these have been deprecated for many years at this point, and the former requires an older, no-longer-supported version of Python.  Instead of trying to hang on to them, I have moved our `.scss` compilation to `sass` - this includes its own middleware functionality, so only the single module is needed.
- `.scss` compilation will now be done 'on demand' - that is, only if the `.scss` files have actually changed.  The compilation will now create an actual `custom.css` file in the repository, and the dashboard calls this file.  Note that compilation can also now only be done on Development, with the `.css` file being updated in the main repository as needed.

Other Changes:
- moved Node version from 16.15.0 to 22 in the docker file
- moved Node version from 16.15.0 to 22 OKD build configuration file [PLEASE LET ME KNOW IF THIS IS INCORRECT - will revert if needed]
- added more details to the main README about how to actually install the NPM modules from scratch (was missing previously for some reason)
- fixed some formatting errors in the OKD README file, and renamed it to `/okd/README.md` so that Github actually picks it up as the file to use as the README in this subdirectory


bjrebel (approved):
Good to update to latest versions, let's give it a go

PR #345, opened 2025-06-30
Update buildconfig.yaml

bjrebel:
Get back to using the FNAL imageregistry


krishmaj (approved):
Thanks for spotting and fixing.

PR #346, opened 2025-07-04
Reorganising All Routes + Component Names Bug Fix

krishmaj:
Reorganising all routes
- a bug was identified which was causing some routes to return an empty object instead of an actual response
- it was originally thought that it was somehow due to how the non-API and API routes were organised into separate files, and therefore registered in the app separately (the bug was only affecting certain API routes)
- merged all API routes into the corresponding non-API routes files ... i.e. all 'component' related routes are now in a single file, etc.
- the 'forms' routes are all API-only, so this file has just been moved from '/routes/api/' to '/routes/'
- all 'users' routes have been moved into '/routes/start.js', since they are all in the '/***' group
- it turns out that the actual reason for the bug was duplicated route signatures ... reworked the 'forms' routes to remove the duplication
- fixed some incorrect returned status codes ... GET routes now all return 200, POST routes now all return 201
- added a reference list of all routes to the 'routes/index.js' file

Small fix to component names
- DWA and DWAPDB component names (and DUNE PIDs) should use the user-defined production numbers, not the DB-assigned type record numbers


bjrebel (approved):
Good changes to test

PR #349, opened 2025-07-20
Miscellaneous Requested Changes

krishmaj:
Adding new freeform section to APA Assembly workflow
- the 'Cable Conduit Insertion' and 'Installation into ASF' actions can now be done in either order, since the tasks within can (and apparently do) overlap
- had to rewrite the entire .pug file to accommodate multiple freeform sections in a single workflow ... the old code was already too complicated to handle any further changes
- each workflow type now has its own section of code ... some extra code duplication now, but far easier to maintain and change in the future

Changes relating to performing bulk QA actions via M2M
- the M2M script (NOT KEPT IN THIS REPOSITORY) will now first check if each board (CE Adapter, CR, G-Bias or SHV) is conformant, and only perform the QA action if so
- changed the library function to pass the component's 'data' object (containing the 'boardIsConformant' field in the case of these component types) back to the script
- fixed a bug where the 'boardIsConformant' field was not being saved if these components were created via a batch

Expanding search for components by type and location
- added new search filters for board conformance (only used for populated boards) and QA check status (only used for populated boards and cable harnesses)
- changed API route and library function accordingly, and updated the index of routes
- added bespoke results layout for populated boards and cable harnesses, so that the QA check status can be displayed only for these component types
- updated search descriptions page

Bug fix for incorrect redirection when performing action on an unspecified component
- forgot to change the URL to the new route in the Formio ComponentUUID component (made the change in the .pug file previously)

### DIFF
--- a/app/lib/Search_OtherComponents.js
+++ b/app/lib/Search_OtherComponents.js
@@ -377,6 +377,7 @@ async function componentsByTypeAndNumber(typeFormId, typeRecordNumber) {
       typeRecordNumber: { '$first': '$data.typeRecordNumber' },
       formName: { '$first': '$formName' },
       shortUuid: { '$first': '$shortUuid' },
+      data: { '$first': '$data' },
     },
   });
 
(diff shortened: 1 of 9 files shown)


bjrebel (approved):
Let's give these a try

PR #352, opened 2025-07-23
Bug Fix and Updating M2M Common File

krishmaj:
Bug fix for incorrect route when creating a workflow component
- forgot to update the routes on the workflow information page

Updating M2M Common File (GetListOfActions function)
- the function will now accept an optional 'component UUID' argument - if specified, it will return only the actions of the specified type that have been performed on the specified component
- if the component UUID is not specified (default = empty string), it will return all actions of the specified type across all components (as previously)
- this functionality already exists on the server side of the DB, so no other changes are required

Adding Stephen Sumner to list of personnel authorised to signoff on APA QA (requested by Radosav)


bjrebel (approved):
Looks good to test

PR #354, opened 2025-07-25
Fixes and Requested Changes

krishmaj:
Adding ability for users to delete images from actions
- images can now be removed from an action record through the web interface
- added new section below image gallery on action information pages, where the user can input the number of the image to be removed and then confirm
- new API route and library function, and added route to index list

Fixes and Requested Changes to Component Names
- bug fix for DWA and DWAPDB components ... were previously using 'data.dwaNumber' and 'data.pdbNumber' in their component names, except those fields don't exist - changed to using 'data.typeRecordNumber', which is the correct thing anyway
- all populated board kits and the multi-type shipment will now have the 'data.typeRecordNumber' in their names (will need to run the Admin Utility to update existing components)
- moved the current Admin Utility function for setting component names and DUNE PIDs to a new library file ... expecting to have more such functions in the future, so best to keep them in the same place (updated route to point to new library file as well)

### DIFF
--- a/app/lib/Actions.js
+++ b/app/lib/Actions.js
@@ -189,7 +189,7 @@ async function addImageStrings(actionId, imageStringsArray) {
 
   match_condition.actionId = new ObjectId(match_condition.actionId);
 
-  // Use the MongoDB '$set' operator to directly edit the values of the relevant fields in the action record, and throw an error if the edit fails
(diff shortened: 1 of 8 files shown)


bjrebel (approved):
Let's try it out

PR #356, opened 2025-07-28
NCR Updates and Consolidating Type Record Numbers

krishmaj:
Consolidating frame and APA numbers with type record numbers
- currently, APA frames have separate 'frame' and 'type record' numbers, and Assembled APAs have separate 'production' and 'type record' numbers
- there is no reason for these to be separate things ... just like with some of the other components, we can use the type record number as the frame or production number
- added new admin utility function to copy values of existing frame and production numbers to the type record number fields in ALL versions of each component
- replaced all calls to the 'data.frameNumber' and 'data.apaNumberAtLocation' fields with 'data.typeRecordNumber' ... this will only take effect after running the admin utility once on all frames and APAs
- (Krish TODO) REMEMBER TO UPDATE THE COMPONENT TYPE FORMS AS SOON AS THESE CHANGES ARE DEPLOYED

Updates to NCR-related code
- new NCR types were recently added to the action type form (with corresponding information tables), and some descriptions were expanded
- made corresponding changes in the APA Executive Summary and Search for NCRs codes (library functions and interface pages)


bjrebel (approved):

PR #358, opened 2025-07-30
Requested Miscellaneous Changes

krishmaj:
Displaying CE Adapter Board Shipment locations
- for some reason, the origin and destination locations of CE Adapter Board Shipments were not being displayed on the component info pages
- this is the only shipment-type component where they weren't ... every other one had them displayed

Changes relating to non-unique Cable Harness typeRecordNumbers
- the 'display bulk QR codes' page will now account for the situation where each typeRecordNumber can correspond to multiple components
- the 'Search for Components by Type and Type Record Number' will show a list of matching component records when searching for cable harnesses with non-unique typeRecordNumbers ... for any other component type that matches multiple components, an error message will still be shown as before


bjrebel (approved):
Let's test this out

PR #360, opened 2025-08-03
Overhaul of Search and Listing Interface Pages

krishmaj:
Changes to listing pages
- currently, the record listing pages for all and single-type records share the same template ... makes it difficult to show relevant information for individual record types, without the code and layouts becoming very complicated
- have now split the templates for all and single-type records into separate files
- layouts for actions and workflows are basically as they were ... some fixes and tweaks to column order and sizes, but essentially the same information is still being displayed for all records and across all types
- added some bespoke information to the listing pages for specific component types ... frames, APAs (requested by Ged), shipment-types and board-types

Adding ASF component naming and DUNE PID construction
- new ASF components will be given names of the format 'ASF 00001-UK', etc. with the user providing the typeRecordNumber and location at component creation
- DUNE PIDs are not assigned, but a template is still set to be consistent with other component types
- added similar code to the admin utility for renaming components (for completeness ... there are no ASF components currently in the DB)

Overhauling search interface pages
- moved search descriptions to be in collapsible sections at the top of each corresponding search page, and removed .pug file and route previously associated with the Search Descriptions page
- most search pages now display results in columns, to reduce the length of the page when lots of results are returned
- improved and made more consistent the layouts and wording of user inputs
- fixed some out-of-date URLs in the 'ActionID' and 'WorkflowID' Formio component code ... don't think it actually matters, since the redirection is handled by the interface, but just in case
- no changes to any search functionality ... all changes are purely on the client-side interface


bjrebel (approved):
Let's try it out on staging

PR #362, opened 2025-08-06
Miscellaneous Follow-Up and Requested Changes

krishmaj:
Adding new personnel list for the CERN Compliance Office

More fixes and tweaks to Search interface pages
- finally figured out how to keep table column widths fixed, regardless of the contents of individual cells
- adding new .css styles for 'results-table' (derivative of the standard Bootstrap 'table') and derived 'th' and 'td'
- any search results that show tables will now use the new 'results-table' styling ... fixed column widths, less padding
- made the default displayed texts for filter variables more consistent and indicative of what the filter is (i.e. 'any' --> 'any disposition')
- fixed some additional issues that I spotted ... split summaries from main results, fixed some incorrect links

Small fix to Search for Components by Type and Location ... when searching for cable harnesses, the side will now be displayed along with the type record number

Updates to M2M files
- updating the common function for retrieving components via type form ID and type record number to now return a list of components (since cable harnesses can have non-unique type record numbers)
- updating the same function to use the 'search for components via type record number' API route ... for some reason we have two routes that do the same thing, but I want to keep this one since it is also used for the interface search - will eventually remove the other routes once I'm sure nobody else is using it
- updated the 'GetRecords' template and README files

### DIFF
--- a/app/lib/Search_OtherComponents.js
+++ b/app/lib/Search_OtherComponents.js
@@ -639,12 +639,21 @@ async function componentsByTypeAndLocation(typeFormId, location, acceptanceStatu
       }
 
       if ((qaChecksStatus === 'any') || ((qaChecksStatus === 'passed') && (qaChecksPassed == true)) || (qaChecksStatus === 'failed') && (qaChecksPassed == false)) {
-        cleanedResults.push({
-          'componentUuid': result.componentUuid,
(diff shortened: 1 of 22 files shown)


bjrebel (approved):
All seem good to test on staging

PR #364, opened 2025-08-07
Bug fix for Component Search by Type and Location ...

krishmaj:
- moved the matching against populated board conformance to be after selecting the latest version of each record
- previously this matching was done along with those against component type and location, but this was wrong - could (and did) result in the wrong version being selected


bjrebel (approved):
Let's give it a go

PR #366, opened 2025-08-14
Requested reorganisation of personnel lists ...

krishmaj:
- added new lists of D-band personnel for frame prep, winding and post-production ... these will be used specifically for signoffs and QA checks during APA Assembly
- removed tension control and grounding mesh panel intake lists ... these are identical to the winder maintenance and (new) Frame Prep D-band lists respectively
- added some comments in the code to clarify exactly where each list is used (still need to figure this out for the UW-related lists)
- updated API routes accordingly ... will update the action type forms once these changes are on Production


bjrebel (approved):
Changes look fine to me

PR #370, opened 2025-08-14
Quick Fixes for APA Executive Summary

krishmaj:
The 'shipping preparation' signoff has been missing since we removed the QA Check specifically for that section of assembly ... have now tied it to the 'Installation into ASF and Close Up' action.  Tested on Staging - the 'close up' signoff from that action is now shown correctly ... at least for APAs 30 and 31 (which are the immediate focus right now).


bjrebel (approved):
tested on staging

PR #371, opened 2025-08-28
Miscellaneous Requests

krishmaj:
Adding Mark Farrell to list of Winding D-Band personnel

Follow-up Changes relating to ASFs
- ASFs will now be shown in (reverse) numerical order on the single-type component listing page ... similar to APA Frames and Assembled APAs
- added ASFs and some other component types to the available selection when searching by type and type record number
- did NOT add shipments, kits, batches or wire bobbins, since these component types do not use their type record numbers for primary identification

Changes to QR Code Displays
- any QR code with a top label will now show both the component name and the component UUID
- this currently applies only when viewing bulk QR codes or when viewing the page of QR codes for a single shipment-type component
- removed the unused 'label' argument from 'no-label' QR codes ... was still being passed for some reason, even though it was not being used
- renamed the internal mixin variables to be clearer - text --> URL, desc --> label

Fixes for Component Location Tracking
- when a component's location is updated, it will now be done for ALL versions of the component ... not just the latest one
- added a new Admin Utility function to update all versions of a single component to have the same reception information as the latest version
- corresponding change on the Admin Utility interface page to reinstate all component types for selection

Major reworking of APA Executive Summaries
- changed how NCRs are grouped - now by disposition (and for 'use-as-is', subgrouped by wires vs. non-wires)
- major cleanup and refresh of backend code ... renaming variables and dictionary keys for better clarity, reorganised flow to match the summary document
- major cleanup and changes to the layout and formatting ... learned a lot of new ways to use and manipulate Formio components since the last time the document layouts were loooked at
- document no longer needs the APA component itself to be passed (any relevant information is now contained in the collated information dictionary) ... removed from route

### DIFF
--- a/app/lib/Admin_Functions.js
+++ b/app/lib/Admin_Functions.js
@@ -3,6 +3,7 @@ const MUUID = require('uuid-mongodb');
 
 const { db } = require('./db');
 const Components = require('./Components');
+const logger = require('../lib/logger');
 const utils = require('./utils');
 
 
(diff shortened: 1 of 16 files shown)


bjrebel (approved):
Tested on staging

PR #373, opened 2025-09-03
More Miscellaneous Changes

krishmaj:
Small final tweaks to APA Executive Summaries
- QC signoff links now identify explicitly which layer is being referred to for winding, soldering and tension measurements
- each NCR group header now shows the count of NCRs in that group, and no message is displayed if there are none (no longer needed)

Fix for listing multi-component-type actions
- some action types can be performed on multiple component types - i.e. populated board QA actions can be performed on all types of populated boards, non-conformance actions can be performed on APAs and meshes
- currently, all listing pages of these action types will show the actions across ALL component types - even though the user gets to those listing pages by clicking on one of the component types ... the expected behaviour would be to list only those actions that have been performed on that component type
- these code changes implement that expected behaviour

Another attempt at fixing the null reception object issue
- removed code from the previous attempt ... clearly didn't work
- finally identified the specific line of code which throws the red-banner error in the interface - when the reception object is 'null', the 'updateLocation' function cannot get the current location
- changed so that the function will now first check if the object is 'null', and only attempt to get the current location if not
- if the object is 'null', the current location will be set to some default string ... it gets changed to the actual value later in the same function anyway
- note that none of this actually fixed why the object could be 'null' in the first place ... still haven't figured that out

### DIFF
--- a/app/lib/Actions.js
+++ b/app/lib/Actions.js
@@ -8,6 +8,7 @@ const dbLock = require('./dbLock');
 const Forms = require('./Forms');
 const permissions = require('./permissions');
 const utils = require('./utils');
+const logger = require('./logger');
 
 // Declare a list of the available 'shipment transport' related action type forms
 // NOTE: this must be the same as the equivalent list given in 'static/pages/action_specComponent.js'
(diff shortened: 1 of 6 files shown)


bjrebel (approved):
Let's test them out

PR #375, opened 2025-09-06
Bug Fix for Multi-Kit Shipment Component Location Tracking

krishmaj:
- the component type form for multi-kit shipments was recently changed to allow any number of kits of each type, including ZERO
- previously, there always had to be at least one kit of each type ... the code was written around that assumption
- changed the code so that it now first checks if a kit exists (for each type), and only attempts to retrieve it and set its location if so
- also applies when retrieving the sub-kits for component summary and bulk-QR code interface pages

While making the above changes, also found and fixed some incorrect summary and QR code displays for other component types (previously missed with the recent changes to component names and QR code displays)

Adding component name to workflow information page
- the name of the component on which a workflow is being performed will now be shown in the top section of the workflow information page
- technicians have mentioned that it's easier to expect it there, in line with actions, and easy to overlook that it's also currently displayed at the top of the path table
- also corrected some small typos on workflow information page as well


bjrebel (approved):
Let's test on staging

PR #378, opened 2025-09-19
Changes relating to APA Shipment Reception actions ...

krishmaj:
- added this action type to the ones which allow images to be added to the record
- new personnel list for signing off on APA shipment reception (dictionary, route and added to routes index)


bjrebel (approved):
All changes look reasonable to me.

PR #380, opened 2025-10-01
Reorganising M2M software across multiple repositories ...

krishmaj:
- so that I don't have to maintain THREE different instances of the M2M software, I'm moving everything into my own Github repo
- removed the M2M scripts and subdirectory ... the one in this repository was always just a backup anyway
- added DSM to the central dictionary of locations - for some reason it was there in the M2M's dictionary but not here, so just keeping things consistent
- added new API route to the search for board shipments by board UUID (library function already exists)


bjrebel (approved):

PR #384, opened 2025-10-28
Adding ability to search for Factory Board Rejection actions ...

krishmaj:
...  that have been performed on a specified board

Search route already exists for Board Installation and Winding actions ... added equivalent new library function for Factory Board Rejection actions


bjrebel (approved):
Changes look good to me.

PR #386, opened 2025-11-04
Adding ability to upload shocklogger plots images to APA Shipment Reception actions ...

krishmaj:
- extended existing image-upload functions to allow for different 'types' of images ... 'general' (what already existed) and 'shocklogger' (new)
- only one shocklogger plots image can be present at a time (new upload will overwrite existing) ... unlike general images (new upload is added to array of images)
- shocklogger plots image is displayed separately from general images (since both can in principle be present in a single APA Shipment Reception action) ... gallery of general images remains as it is


bjrebel (approved):
looks ready to test on staging

PR #388, opened 2025-11-07
Adding framework for HWDB APA data output ...

krishmaj:
- the DUNE HWDB will eventually require data from the APA DB relating to the pairs of Assembled APAs that form each doublet in the detector
- added new library function to collate data about two specified APAs and return it in a preset JSON format (still to be determined)
- added new interface and API routes, and updated routes index
- new interface page, linked in the sidebar - the user will specify the UUIDs of two APAs, and the DB will return the preset JSON formatted information
- users may also copy each APA's returned information to their device's clipboard, or download the information directly as a JSON file

### DIFF
--- a/app/lib/Components_CollateInfo.js
+++ b/app/lib/Components_CollateInfo.js
@@ -69,8 +69,8 @@ const dictionary_ncrTypes = {
 };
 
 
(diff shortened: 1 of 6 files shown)


bjrebel (approved):
Looks ready to test

PR #390, opened 2025-11-14
Changes to Geometry Board Location Tracking

krishmaj:
Geometry boards can now be 'located' as 'rejected' - making it easier to directly track and count which ones have been rejected and removed from use, without needing to indirectly go into the Board Factory Rejection actions.

- added new 'rejected' location to the central dictionary of locations, as well as to the list of available locations in the 'Search for Components by Type and Location'
- when submitting a new or edited 'Factory Board Rejection' action where the disposition is set to 'rejected', the target board's 'location' will now be set to 'rejected', with the 'date' being that of the submission, and the 'detail' being a string containing information about where the rejection occurred and why
- when submitting a new or edited 'Factory Board Rejection' action where the disposition is set to anything else, the target board's 'location' will be set to wherever the rejection action was performed, with the 'date' being that of the submission, and the 'detail' being an empty string
- the combination of 'location' and 'detail' will be shown on the component information page of any rejected boards and the single-type component listing page, in the same way as any other location is currently

- removed 'board acceptance' filter from the 'Search for Components by Type and Location' ... that filter was for accepted vs. rejected boards, which is now done by selecting the 'rejected' location instead
- changed the server-side function for the 'Search for Components by Type and Part Number' to directly match to a geometry board's location if the 'board acceptance' filter is set to 'rejected'
- changed the server-side M2M function for getting 'Board Rejection Counts by Type and Location' to directly match to a geometry board's location

- new Admin Utility to update board locations ... checks if the LATEST VERSION of the MOST RECENT board rejection action has a disposition of 'rejected' - if so, set the board location as per the same rules as for future rejection actions

### DIFF
--- a/app/lib/Actions.js
+++ b/app/lib/Actions.js
@@ -185,6 +185,7 @@ async function save(input, req) {
       else if (newRecord.data.reasonForRejectionFromInventory.toothStripNotProperlyAttached) { rejectionReason = 'Misattached Tooth Strip' }
       else if (newRecord.data.reasonForRejectionFromInventory.qrCodeIssue) { rejectionReason = 'QR Code Issue' }
       else if (newRecord.data.reasonForRejectionFromInventory.installationCausedDamage) { rejectionReason = 'Installation Damage' }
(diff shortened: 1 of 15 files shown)


bjrebel (approved):
Changes are ready to test on staging

PR #391, opened 2025-11-15
Bug Fixes for Recent Changes ...

krishmaj:
- board acceptance status filter was not being applied correctly when set to 'accepted' in the Search for Components by Type and Part Number
- needed to project the 'actionId' field in the recently added Admin Utility ... otherwise the subsequent grouping stage wouldn't work (since it groups by actionId)
- added 'directConnection' argument to internal DB URL for Development instance ... apparently NodeJS 17+ works better with this set, to avoid localhost server timeouts (shouldn't make any difference to OKD deployments)

Also changed the new Admin Utility to use the already-existing 'updateLocations' component library function ... this is a safer option than doing it explicitly in the utility, since the function has additional safety and sanity checks built in already.

### DIFF
--- a/app/lib/Admin_Functions.js
+++ b/app/lib/Admin_Functions.js
@@ -269,9 +269,10 @@ async function updateComponentLocations_viaActions(actionTypeFormId) {
   // Project only the fields that will be needed for the later steps
   aggregation_stages.push({
     $project: {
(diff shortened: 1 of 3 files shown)


bjrebel (approved):
Looks ready to test

PR #393, opened 2025-11-21
Update to admin utility due to mismatched field names ...

krishmaj:
- some older Board Factory Rejection actions have the step failure field name given as 'stepBetweenBoardAndStripTooLarge', instead of 'step' as it is in newer (and for some reason, very old) actions
- similar problem with the 'other' field ... sometimes recorded as 'otherProvideDetailsInTheCommentsBoxOnTheRight'
- the existing admin utility for updating the locations of rejected boards only accounted for 'step' and 'other', not the alternatives
- have also ignored any other rejection reasons ... no need to unnecessarily update and alter those records, since they aren't affected


bjrebel (approved):
let's give it a go

PR #394, opened 2025-11-22
Adding exception to Admin Utility for US APA 1 boards ...

krishmaj:
- these boards were previously rejected in 2024 via rejection actions with a reason of 'other' - and so they would be caught by the utility and have their reception details set to 'Unspecified Reason'
- however, Brian recently changed all such boards to have reception detail of 'Chicago - Removed from APA', which is the correct one ... don't want the utility to overwrite that


bjrebel (approved):
Let's try it out

PR #395, opened 2025-12-13
Miscellaneous Changes

krishmaj:
Removing unused component route
- the '/json/component/{typeFormId}/{typeRecordNumber}' route is no longer used - was previously only used by the 'GetComponent_byTypeRecordNumber' common M2M function, but that now uses the '/json/search/componentsByTypeAndNumber/{typeFormId}/{typeRecordNumber}' route
- the latter is already used by both the API and interface for searching for a component by type and record number, so it should definitely be kept ... no point in having another route that does the same thing only for the M2M side

Small change to APA Executive Summaries
- changed 'bad solders' to 'reworked solders', both in the collation function and in the summary document itself
- based on email discussion with Eric on 2025-12-04

Removing unneeded imports from library files
- actually spotted these a while ago, but kept on forgetting to clean them out

Stronger security checks on certain actions
- applicable to Completed APA QA Checklist, Assembled APA QA Check and Mesh Panel Installation actions
- when submitting these action types, the DB will now compare the name of the submitter (from their Auth0 account) with the name given in the action's QA signoff field
- if they do not match, the action will NOT be submitted - and an error will be displayed (Mesh Installation actions are checked when being completed, others are always checked regardless of completion)
- this should prevent such actions from being submitted and completed on behalf of someone other than the submitter
- also fixed a small typo in one of the technicians' displayed names (cannot fix the dictionary key, since that would require finding and correcting all records where the key was used ... which we don't actually know)

Fix for older Tooth Strip Attachment actions
- the older tooth strip attachment actions are missing two QC checks - probably added relatively recently, but the records were not updated
- this causes them to be shown as 'failed' in the geometry board history table - since this checks for the simultaneous existence and passing of all 6 expected QC checks
- changed component route so that it now checks for existence and passing separately ... and will now show a partial pass if any checks are missing (but the remaining ones all pass)
- similarly, some older visual inspection actions have an empty string for the disposition - route now also accounts for this

### DIFF
--- a/app/lib/Actions.js
+++ b/app/lib/Actions.js
@@ -42,11 +42,24 @@ async function save(input, req) {
 
   if (!typeForm) throw new Error(`Actions:save() - the specified 'input.typeFormId' (${input.typeFormId}) does not match a known action type form!`);
 
-  // Check that the person performing or editing the action is permitted to do so ... this is an additional security check for specific action types, beyond a simple global 'permissions' check
-  // It is designed to make sure that these actions are not being performed by the logged-in user on behalf of someone else, without the latter's knowledge or permission 
+  // Some action types should be submitted only by the APA Factory Leads - these are typically the most important and/or highest level QA checks and signoffs
+  // Check that the submitter (i.e. the currently logged-in user) is the same as the person who's name is being used for the QA signoff ... if not, do not allow the action to be submitted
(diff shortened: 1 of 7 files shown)


bjrebel (approved):
Let's try them out.

PR #397, opened 2025-12-20
More Miscellaneous Changes

krishmaj:
Removing now unused visual inspection dispositions from search
- following on from changes made in previous commits

Adding NCR title to interface display
- when listing NCR actions (on both APAs and mesh panels), the NCR title will now be displayed
- other action types remain unchanged - with an empty column displayed
- code framework can be extended to show additional information for any other action types in the future (as with component types)
- also adjusted displayed dates and times to international standard and for better display (shortened month strings)

Fixes for bugged NCR actions
- for some reason, Formio can sometimes (but not always) write an incomplete entry to any number of the wire data grids in NCR actions
- might be related to whether the user selects and deselects the corresponding checkbox during data input, but not completely sure of the cause
- applied two changes to the code in order to mitigate the resulting crash when attempting to set up the APA Executive Summary
- when saving the NCR action, the DB will now check to see if there are any incomplete entries in the grids, and correct them if found
- when setting up the Exec. Summary, the DB will now check to see if the 'wireLayer' field exists in the wire grids, and only attempt to display them if so

### DIFF
--- a/app/lib/Actions.js
+++ b/app/lib/Actions.js
@@ -379,6 +379,7 @@ async function list(match_condition, options) {
       componentUuid: true,
       workflowId: true,
       validity: true,
+      data: true,
     }
   })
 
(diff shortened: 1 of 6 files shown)


bjrebel (approved):
Let's test on staging

PR #399, opened 2026-01-03
New Admin Utility for Cleaning Up Component Records

krishmaj:
Ready for Use
- we used to include secondary identifiers (i.e. part numbers, under various names) for components of certain types
- these are no longer needed, since the component 'typeRecordNumber' field is now used universally as a secondary identifier (together with the type form ID or name)
- new admin utility function to remove the no-longer-used secondary identifiers, with associated changes to interface page and route

Work in Progress
- also added a new admin utility function that will (in principle) add some more component information to action records ... name, type form ID and type form name
- this function is not currently working properly ... likely because it queries every single action record in the collection, so is probably overloading memory - works fine with smaller numbers of actions ... need to look into it in more detail

### DIFF
--- a/app/lib/Admin_Functions.js
+++ b/app/lib/Admin_Functions.js
@@ -1,4 +1,3 @@
-const Binary = require('mongodb').Binary;
 const MUUID = require('uuid-mongodb');
 
 const Actions = require('./Actions');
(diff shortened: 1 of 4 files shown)


bjrebel (approved):
Let's give these a go

PR #400, opened 2026-01-11
Fixing the admin utility for adding component info to actions ...

krishmaj:
- admin utility will now retrieve and change only the records corresponding to the selected action types - selected via the component type (or group of types)
- the same component information will now be added to all actions when they are first performed or later edited
- fixed a bug in the previous admin utility ... changed one of the interface options, but forgot to change the backend code accordingly


bjrebel (approved):
Let's give it a go

PR #401, opened 2026-01-13
Splitting admin utility usage to get around memory overload ...

krishmaj:
- attempting to change all actions related to geometry boards is causing a memory overload on Staging
- changed so that each action type is now called individually


bjrebel (approved):
Let's give it a shot.

PR #402, opened 2026-01-13
Cleaning Up Component Records and Additions to Action Records

krishmaj:
New admin utility to do the following:
- remove unneeded fields from component records of specific component types
- add more component information to all action records

Further improvements to the code - to take advantage of the additional component information in action records - will follow in a later commit.


bjrebel (approved):
tested on staging

PR #403, opened 2026-01-14
Improving action retrieval speeds using new record fields ...

krishmaj:
- there are a handful of places in the interface where one or more action records are retrieved, and currently the associated component record must also be retrieved in order to access and display the component name, type, etc.
- this is noticeably slow, particularly for bulk retrievals (action listing pages)
- since the component information is now found within the action record, there is no longer any need for the additional component record retrieval - code has been changed accondingly ... action listing is significantly faster now


bjrebel (approved):
Tested on staging

PR #404, opened 2026-01-14
Requested additions to Winder Maintenance Verification personnel list

krishmaj:
Daresbury Factory Leads have requested that the Winding D-Band personnel be able to sign off on winder maintenance in Winding actions


bjrebel (approved):
simple change to a list, nothing to worry about here

PR #406, opened 2026-01-15
New personnel list for tension control verification ...

krishmaj:
- requested that the tension control and winder maintenance personnel lists be separated - recent additions for the latter should not be in the former
- new dictionary and associated route ... will make the corresponding changes on winding action type forms once the code changes have been deployed


bjrebel (approved):
looks like a simple change

PR #410, opened 2026-01-23
Additions to component location tracking ...

krishmaj:
- added assignment of initial locations for APA Shipping Frames (frame location) and Yokes ('Wisconsin') - now only batch-type components have no initial location set
- ASF location can now be changed (to 'In Transit') when referencing it in an APA Shipment, and again (to the reception destination) when receiving an APA Shipment via its reception action
- added additional geometry board and mesh panel actions that can change the component location (all of these action types contain an 'action location' field) ... this will help to fill in the gaps between shipments of the components
- added 'US Warehouse' as a new entry in the central list of locations ... will be used in the future as the destination for shipping yokes (not yet implemented)


bjrebel (approved):
Changes look fine to me.

PR #411, opened 2026-01-28
Location tracking for frame and yoke shipments ...

krishmaj:
- completing location tracking for frame shipments - shipments and frames will now have their locations set to 'in transit' when the shipment is created, and to [destination] when the shipment is received
- adding location tracking for yoke shipments (new component type) - shipments and yokes will now have their locations set to 'in transit' when the shipment is created, and to [destination] when the shipment is received
- adding location tracking for CE adapter board shipment transport - when the shipment transport action is performed, the shipment and its boards will have their locations set to 'in transit' (the locations are not changed when the shipment is created)
- changed 'US Warehouse' location to 'South Dakota Warehouse' as requested, for less ambiguity
- added Ste Sumner to frame-prep D-Band personnel list as requested


bjrebel (approved):
Let's give it a shot

PR #415, opened 2026-03-03
Small rework of geometry board rejection by removal from APA

krishmaj:
- currently, geometry boards cannot be 'rejected' via removal from an APA, since the DB safeguards against changing the location of already-installed boards
- added a new function that allows boards to be removed and rejected ... separate from the existing 'updateLocation' function
- the new function will only be called when performing a 'Board Factory Rejection' action AND when the disposition is 'rejected' AND when the rejection reason is 'Removed from APA'


bjrebel (approved):
let's give it a go

PR #418, opened 2026-03-06
Reworking HWDB Information Dump functionality

krishmaj:
- after discussion with Brian, we realised that the HWDB info dump needs to (eventually) work for other component types, but doesn't need to output simultaneously for more than one at a time
- interface page now only needs one component UUID specified ... fixed surrounding layout accordingly
- changed route and library function accordingly - framework is now in place to extend functionality to other component types as needed in the future
- existing functionality for Assembled APAs remains unchanged

Also making a small change to allow APA Cold Test actions to include images

### DIFF
--- a/app/lib/Components_CollateInfo.js
+++ b/app/lib/Components_CollateInfo.js
@@ -741,20 +741,24 @@ async function forExecSummary(componentUUID) {
 }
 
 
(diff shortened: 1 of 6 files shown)


bjrebel (approved):
let's try it out on staging

PR #420, opened 2026-03-10
Changing APA Frame Shipment component type form ID ...

krishmaj:
- the 'APA Frame Shipment' component type currently has an form ID and name of 'Frame Shipment'
- changed to 'APA Frame Shipment' to make it consistent with the actual APA Frames
- change already made to type form (actually just made a new one with the corrected name, since I cannot change an existing type form's ID) ... simple to do for this one, since there are no components of this type


bjrebel (approved):
Let's give the changes a test on staging

PR #423, opened 2026-03-27
Request from Radosav for additional personnel names ...

krishmaj:
- added Radosav to the list of D-Band post-production personnel
- added a generic 'CERN Staff' entry to the list of all technicians - to be used for when APAs are being packaged at CERN for transport elsewhere


bjrebel (approved):
simple change

PR #426, opened 2026-04-08
Component Type Form ID and Name Consistency Changes (PR 1 of 2)

krishmaj:
Adding new Daresbury technician to central list

Component Type Form ID and Name Consistency Changes (PR 1 of 2)
- there are 4 component types which have type form IDs and/or names which are inconsistent with all other types - want to clean these up ... 3 to be dealt with in this PR
- Grounding Mesh Shipment (GroundingMeshShipment) --> Grounding Mesh Panel Shipment (GroundingMeshPanelShipment)
- Wire Bobbin (wire_bobbin) --> Wire Bobbin (WireBobbin) [ID change only]
- Board Shipment (BoardShipment) --> Geometry Board Shipment (GeometryBoardShipment)
- changed all instances of the formers in the code to the latters ... i.e. where an interface page layout is tailored to a specific component type, or where specific information is to be retrieved depending on the component type
- added a new administrator utility function to directly change the 'formId' and 'formName' fields of all relevant component records
- corresponding changes to the two search functions that apply to geometry board shipments (library functions, routes and interface page file names)

Note that for each component type, a new type form will need to be created through the interface, and any connected existing action type forms then edited to point to it (the old component type form can then be trashed)

### DIFF
--- a/app/lib/Admin_Functions.js
+++ b/app/lib/Admin_Functions.js
@@ -18,6 +18,9 @@ async function cleanComponentTypeFormIds(typeFormId) {
   } else if (typeFormId === 'wire_bobbin') {
     newTypeFormId = 'WireBobbin';
     newTypeFormName = 'Wire Bobbin';
+  } else if (typeFormId === 'BoardShipment') {
+    newTypeFormId = 'GeometryBoardShipment';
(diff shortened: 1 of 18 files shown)


bjrebel (approved):

PR #428, opened 2026-04-11
Component Type Form ID and Name Consistency Changes (Part 4) ...

krishmaj:
- continuing with 'APA Shipment' ... renamed to 'Assembled APA Shipment'
- changed all instances of the former to the latter
- additional code in the administrator utility function to change the 'path[0].formName' field in 'APA Post Production' workflows to be the new type form name
- changed code for setting component names at creation to now use the updated type form names (previously, some types were using hardcoded naming because their type form names were different from desired ... not anymore)


bjrebel (approved):
Let's test on staging

PR #429, opened 2026-04-17
Miscellaneous Changes and Fixes

krishmaj:
Changes to FD-HD Technical Coordinator signoffs ...
- after discussion with Brian, realised that the FD-HD technical coordinator signoff is actually the final one for 'Completed APA QA Checklist' actions ... that is the one that should be checked against the submitter
- added new dictionary of FD-HD technical coordinators (just Eric at the moment), and corresponding route
- the 'Completed APA QA Checklist' action type should now use the new dictionary, instead of the built-in list of names
- if attempting to be completed, the action will now check the inputted FD-HD technical coordinator against the new dictionary ... submission rejected if there is no match
- small clean up of displayed errors and unused dictionaries

Fixing outdated component names in actions ...
- actions performed on Assembled APA Shipments still show the old-style component names ('APA Shipment XXX') ... forgot to update these when I updated the other fields in the previous commit
- set up administrator utility to update these component names

### DIFF
--- a/app/lib/Actions.js
+++ b/app/lib/Actions.js
@@ -39,24 +39,24 @@ async function save(input, req) {
 
   if (!typeForm) throw new Error(`Actions:save() - the specified 'input.typeFormId' (${input.typeFormId}) does not match a known action type form!`);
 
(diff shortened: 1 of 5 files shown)


bjrebel (approved):
looks ready to test on staging

PR #431, opened 2026-04-29
Reverting admin utility to fix more old component type names and IDs ...

krishmaj:
- some components of the old types have recently been created accidentally
- changing the admin utility back to the one which can be used to update the component types
- have reminded the users in question of the change


bjrebel (approved):
Let's try it on staging

PR #432, opened 2026-05-06
Additional change for the previous commit ...

krishmaj:
- forgot to change the component names and DUNE PIDs
- since those are unique to each component, had to take the admin utility back even further to the function that was previously used to set the individual component names ... now also sets the correct type form names and IDs


bjrebel (approved):
This is why we test on staging

PR #433, opened 2026-05-07
Changes to Admin Utility to Fix Incorrect Records

krishmaj:
Some geometry board shipment and wire bobbin component records have been recently created using the old component type forms ... the former due to an un-updated M2M script, and the latter due to the old form not being trashed.

The administrator utility has been reverted to an older function in order to change the records' type form IDs and names, and fill in the currently missing component names and DUNE PIDs.


bjrebel (approved):
tested on staging

bjrebel (approved):
Tested on staging.

PR #436, opened 2026-05-15
Fixing Component Record formId and formName fields (Part 1) ...

krishmaj:
- component records have had 'formId' and 'formName' fields from the very start of the APA DB, whereas the same fields in action and workflow records have always been called 'typeFormId' and 'typeFormName'
- absolutely no reason why these fields should be different depending on entity type ... they mean the same thing, and are assigned in the same way - literally the only difference is the field name
- new administrator utility function to add 'typeFormId' and 'typeFormName' fields to all component records, with values copied from the existing 'formId' and 'formName' fields respectively
- new components will now have the 'typeFormId' and 'typeFormName' fields added, in addition to the 'formId' and 'formName' ones
- changed action and workflow records to pull the type form ID directly from the type form itself, instead of the submitted input ... keeps them consistent with (new) component records, and is more reliable in case the submitted input is malformed

### DIFF
--- a/app/lib/Actions.js
+++ b/app/lib/Actions.js
@@ -79,7 +79,7 @@ async function save(input, req) {
 
   newRecord.recordType = 'action';
   newRecord.actionId = new ObjectId(input.actionId);
-  newRecord.typeFormId = input.typeFormId;
+  newRecord.typeFormId = typeForm.formId;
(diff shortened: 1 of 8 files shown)


bjrebel (approved):
Let's test on staging

PR #438, opened 2026-05-21
Fixing Component Record formId and formName fields (Part 2) ...

krishmaj:
- changing code to use the new 'typeFormId' and 'typeFormName' component record fields instead of the old 'formId' and 'formName' ones
- removed any places where the latter were still being assigned or set ... all new records from now on will only have the former present
- note that 'formId' and 'formName' are still the relevant field names for the actual type form records themselves ... might change those too in the future, but for now they can stay as they are

Additional Administrator Utility for Batch Components
- added new administrator utility function to change 'subComponent_formId' field name to 'subComponent_typeFormId'
- actually just adds the new field with the same value as the old one, like with 'formId' and 'formName' previously
- field name change was already made to the type forms via the interface, so new batches since last week should be fine - this will only apply to older ones


bjrebel (approved):
Let's try it out

PR #440, opened 2026-06-03
Fixing Component Record formId and formName fields (Part 3) ...

krishmaj:
- set administrator utility to remove the now unused 'formId' and 'formName' fields from all component records
- new administrator utility to update Wire Bobbin component names to include the manufacturer, and display Lot Number vs. RM Tag accordingly (naming format has been agreed with Dan Salisbury at Daresbury)
- updated component naming code to do the same for new Wire Bobbins in the future
- corresponding new common dictionary of wire spool manufacturers, with associated route so it can be used in the Wire Bobbin component type form once deployed, and change APA Exec. Summaries to use the new dictionary

### DIFF
--- a/app/lib/Admin_Functions.js
+++ b/app/lib/Admin_Functions.js
@@ -49,7 +49,69 @@ async function fixFields_batchComponents() {
 }
 
 
+async function removeFields_allComponents(currentTypeFormId) {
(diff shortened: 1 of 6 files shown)


bjrebel (approved):
Looks ready for testing

PR #443, opened 2026-06-16
Flattening component location information (Part 1) ...

krishmaj:
- there is no longer any need to have a specific 'reception' object to hold component location information - most component types aren't even received anywhere, so the current setup is also a bit misleading
- additionally, having the location information stored within an object instead of as top-level fields makes it needlessly complicated to access (the object has to be checked first, then the fields, etc.)
- new components will now have location information saved directly into the top level of the record (currently alongside the reception object)
- new admin utility to copy existing reception information into the same top level fields

### DIFF
--- a/app/lib/Admin_Functions.js
+++ b/app/lib/Admin_Functions.js
@@ -10,108 +10,43 @@ const utils = require('./utils');
 const Workflows = require('./Workflows');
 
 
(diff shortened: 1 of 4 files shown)


bjrebel (approved):
Tested on staging and ready to go

PR #444, opened 2026-06-20
Flattening component location information (Part 2) ...

krishmaj:
- changed code to use the new location, dateAtLocation and locationDetail fields
- added administrator utility function to remove the now unused 'reception' object from all component records


bjrebel (approved):
Let's try it out on staging

PR #446, opened 2026-07-13
New Personnel List and Frame Name Bug Fix

krishmaj:
New personnel list for Conduit Insertion APA Assembly actions
- Dave Brown is now allowed to sign off on conduit insertion checks
- currently, the insertion checks use the D-Band Post Production personnel list - but so do a number of other checks, which Dave is not supposed to be able to sign off on
- made a new personnel list specifically for conduit insertion checks, which is just the D-Band Post Production one plus Dave

Bug fix for new APA Frame production location
- frame component name will now be correctly set for EVT-produced frames (previously assumed that all UK frames came from DSM)
- added EVT to the global dictionary of locations

### DIFF
--- a/app/lib/Components.js
+++ b/app/lib/Components.js
@@ -93,7 +93,7 @@ async function save(input, req) {
     if (newRecord.typeFormId === 'APAFrame') {
       let pidSuffix = '';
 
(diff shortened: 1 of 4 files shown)


bjrebel (approved):
Looks ready to test.

PR #448, opened 2026-07-15
New personnel list for yoke signoff ...

krishmaj:
- actually just repurposed an existing, but empty and unused, list of yoke technicians at UW


bjrebel (approved):
Simple change, let's test it on staging

PR #450, opened 2026-07-20
Improving connections between APAs, ASFs and APA Shipments ...

krishmaj:
- each Assembled APA component info page will now show a link to the APA Post Production workflow whose APA Shipment component includes the APA in question (if the APA is not yet associated with a shipment component, no link will be shown)
- similarly, each ASF component info page will now show a link to the APA Post Production workflow whose APA Shipment component includes the ASF in question (and again, if the ASF is not yet associated with a shipment component, no link will be shown)
- the component listing page for ASFs will now show which APA Shipment it is associated with (or a message indicating otherwise, if the ASF is not yet in use)

### DIFF
--- a/app/lib/Search_OtherComponents.js
+++ b/app/lib/Search_OtherComponents.js
@@ -322,6 +322,50 @@ async function apasByProductionLocationAndAssemblyStep(location, assemblyStep) {
 }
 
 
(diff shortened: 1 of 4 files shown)


bjrebel (approved):
Ready to test on staging

PR #451, opened 2026-07-21
Miscellaneous Changes

krishmaj:
New personnel list for populated board shipment signoff
- reusing currently unused 'UW Installation Signoff' personnel list for convenience

Replacing yoke shipments with general installation hardware shipments
- installation hardware includes more than just the yokes, so we need a shipment type that covers everything in one go
- previous 'YokeShipment' type component has been replaced with a new 'InstallationHardwareShipment' one, and similarly with their reception actions
- new shipment type should be displayed fully on the component info and summary pages (instead of a condensed version like other shipment-like components)
- when displaying QR codes for the new component type, use a shortened component name (otherwise it's too long, and the UUID gets pushed out)

Adding new engineering release signoffs to component info pages
- the CE Adapter Board Shipment and Multi-Type Populated Board Shipment component info pages display a condensed version of the type form for convenience
- added the new 'Engineering Release' signoffs to this display ... otherwise what's the point

### DIFF
--- a/app/lib/Actions.js
+++ b/app/lib/Actions.js
@@ -13,7 +13,7 @@ const utils = require('./utils');
 const transport_typeFormIDs = ['APAShipmentTransport', 'CEAdapterBoardTransport'];
 
 // Declare a list of the available 'reception' related action type forms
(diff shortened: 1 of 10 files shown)


bjrebel (approved):
Let's test on staging

PR #453, opened 2026-07-23
Adding new inputs to populated board kit information pages ...

krishmaj:
- populated board kits and shipments now include an Engineering Release input, plus a comments box (mainly to show who submitted them if creating via M2M)
- added these inputs to the bespoke component information pages for these component types (they don't show the entire type form, but instead use a reduced version tailored to each component type)


bjrebel (approved):
Looks ready to test

PR #454, opened 2026-07-23
Recent Requested Changes from Staging

krishmaj:
Improving connections between APAs, ASFs and APA Shipments
- each Assembled APA component info page will now show a link to the APA Post Production workflow whose APA Shipment component includes the APA in question (if the APA is not yet associated with a shipment component, no link will be shown)
- similarly, each ASF component info page will now show a link to the APA Post Production workflow whose APA Shipment component includes the ASF in question (and again, if the ASF is not yet associated with a shipment component, no link will be shown)
- the component listing page for ASFs will now show which APA Shipment it is associated with (or a message indicating otherwise, if the ASF is not yet in use)

New personnel list for populated board shipment signoff
- reusing currently unused 'UW Installation Signoff' personnel list for convenience

Replacing yoke shipments with general installation hardware shipments
- installation hardware includes more than just the yokes, so we need a shipment type that covers everything in one go
- previous 'YokeShipment' type component has been replaced with a new 'InstallationHardwareShipment' one, and similarly with their reception actions
- new shipment type should be displayed fully on the component info and summary pages (instead of a condensed version like other shipment-like components)
- when displaying QR codes for the new component type, use a shortened component name (otherwise it's too long, and the UUID gets pushed out)

Adding new inputs to populated board kit information pages
- populated board kits and shipments now include an Engineering Release input, plus a comments box (mainly to show who submitted them if creating via M2M)
- added these inputs to the bespoke component information pages for these component types (they don't show the entire type form, but instead use a reduced version tailored to each component type)

All changes have been tested on Staging.

### DIFF
--- a/app/lib/Actions.js
+++ b/app/lib/Actions.js
@@ -13,7 +13,7 @@ const utils = require('./utils');
 const transport_typeFormIDs = ['APAShipmentTransport', 'CEAdapterBoardTransport'];
 
 // Declare a list of the available 'reception' related action type forms
-const reception_typeFormIDs = ['APAShipmentReception', 'BoardReception', 'CEAdapterBoardReception', 'DWAComponentShipmentReception', 'FrameShipmentReception', 'GroundingMeshShipmentReception', 'PopulatedBoardKitReception', 'YokeShipmentReception'];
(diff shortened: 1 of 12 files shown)


bjrebel (approved):
All tested on staging

PR #455, opened 2026-07-28
Flattening validity and insertion objects (Part 1) ...

krishmaj:
- of the 6 total fields in the validity and insertion objects, 4 are actually used and 1 of those is a duplicate ... don't really any of them to be nested objects either
- new records will now write only the useful information directly into top-level fields
- changed administrator utility to copy the existing information into new top-level fields in existing records


bjrebel (approved):
Let's test on staging

PR #457, opened 2026-07-29
Flattening validity and insertion objects (Part 2) ...

krishmaj:
- changing backend to use the new top-level fields (added using Administrator Utility)
- administrator utility changed to remove the now unused Validity and Insertion objects from all records


bjrebel (approved):
let's test it on staging

PR #459, opened 2026-07-30
Bug fix for NCRs with non-APA component types ...

krishmaj:
- each of the three (now actually two) component types in an NCR saves the NCR type(s) into a different object
- the code for generating an APA Executive Summary previously only accounted for APA component types ... if it was one of the others, the APA-related object would be null and the summary would fail to generate
- changed the code so that it now tries to retrieve all three objects, and then uses the one that isn't null


bjrebel (approved):
Good to get this bug fixed

